PR #1, opened 2025-09-03
Add example dataset loading functionality

vetschn:
This still needs some thinking. Also we need to somehow provide proper setups, not just the input data.

As we have the same input data for multiple examples but different settings in the `config.toml` (e.g. carbon nanotube with either QTBM / phonons or coulomb Screening), we could maybe have all the small input text files directly in the repo as well and then just load the large raw data on demand. We could perhaps have symlinks to the input data that are dangling until the data is actually loaded. This way we also could ensure that the input files remain compatible with the code version.

The other option would be to have example archives including configs sitting in some other repo and loading specific subfolders. I have not found a way to download only selected subfolders on github, huggingface or zenodo, so probably won't work nicely.

- [ ] Gather all the data
- [ ] Decide how to best set this up
- [ ] Set up all the different examples and document them a little


almaeder:
I believe this should also include the CLI to run the example and remove the need for the run scripts.
The cli would also allow us to fetch a certain example nicely.

almaeder:
"The other option would be to have example archives including configs sitting in some other repo and loading specific subfolders. "
Not a huge fan since the configurations will still be rapidly changing.

almaeder:
We could change the enforcement of a fixed file name.

almaeder:
Tests are expected to fail for quatrex since the dev branch is brocken.

.github/workflows/tests.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Miniconda
         uses: conda-incubator/setup-miniconda@v3
         with:
-          python-version: "3.11"
+          python-version: "3.12"
```
alexnick83: Let's move this to 3.13 since it is supported now on all dependencies. Going forward, and after an initial release, we should make a plan on how many Python versions we will support and for how long.

.github/workflows/tests.yaml
```diff
@@ -33,8 +33,8 @@ jobs:
 
       - name: Run MPI Tests
         run: |
-          NUMBA_DISABLE_JIT=1 mpiexec -np 6 pytest --only-mpi tests/qttools
+          NUMBA_DISABLE_JIT=1 mpiexec -np 6 pytest --only-mpi tests
```
alexnick83: Shall we resume testing with Numba? Or are we planning to have a separate set of tests with it?

environment-dev.yaml
```diff
@@ -3,74 +3,22 @@ channels:
...
-  - toml=0.10.2
-  - tomli=2.0.2
-  - tzdata=2024b
-  - ukkonen=1.0.1
-  - virtualenv=20.26.1
-  - wheel=0.44.0
-  - xz=5.4.6
-  - yaml=0.2.5
-  - zlib=1.2.13
+  - black>=24.8.0
```
alexnick83: Here, we should just provide an exact conda working environment, after we remove modules installed with pip and special flags. We can always provide general dependencies and min/max versions in the setup.

alexnick83 (changes requested):
The general structure is excellent. I pushed a couple of commits to enable the example to run (needs determination of the energy window, which ends up having 14k points), but these should be further refactored. We also need a smaller test that can fit in the runner.

.github/workflows/tests.yaml
vetschn: This still tests all the numba kernels but just turns of the JIT which allows us to get a more accurate coverage rate (jitified kernels were counted as onlz one line)

src/quatrex/cli/main.py
```diff
@@ -0,0 +1,208 @@
...
+ \__, |\__,_|\__,_|\__|_|  \___/_/\_\
+    |_|                              
+
+
+                           version {quatrex.__version__}
+"""
+
+quatrex_cli = typer.Typer(
+    add_completion=False,
+    rich_markup_mode="markdown",
```
vetschn, line 34: Here we'll need to also add `pretty_exceptions_show_locals=False` to avoid crazy long tracebacks

.github/workflows/tests.yaml
alexnick83: I see. To be clear, according to the [documentation](https://numba.readthedocs.io/en/0.61.2/reference/envvars.html#envvar-NUMBA_DISABLE_JIT), setting the flag makes it so the original Python code is called. It may also make sense to run (just) the relevant tests a second time, with the JIT enabled, to ensure that the Numba-optimized code is actually running and returning the correct results.

awinka (approved):
LGTM

almaeder:
It is not yet so great how the files are handled with the manifest and the naming that I did.

alexnick83 (approved):

PR #5, opened 2025-09-05
[vincent-maillou/qttools#161] [vincent-maillou/quatrex#27] Add wave function solver (QTBM)

### LINKED ISSUE #102: QTBM
<a href="https://github.com/maurodossena"></img></a> **Issue by [maurodossena](https://github.com/maurodossena)**
_Tuesday Feb 25, 2025 at 09:40 GMT_
_Originally opened as https://github.com/vincent-maillou/quatrex/pull/27_

----

Added QTBM and some examples

----

_**[maurodossena](https://github.com/maurodossena)** included the following code: https://github.com/vincent-maillou/quatrex/pull/27/commits_

### LINKED ISSUE #175: GPU Eig: _eig_cupy does not return anything

almaeder:
**Issue by [vetschn](https://github.com/vetschn)**
_Originally opened as [https://github.com/vincent-maillou/qttools/pull/161](https://github.com/vincent-maillou/qttools/pull/161)_

This PR was originally in qttools repository (branch: add_cuDSS).

Original PR description:

----

This add direct sparse solvers for QTBM.

----

**Issue by [maurodossena](https://github.com/maurodossena)**
_Originally opened as [https://github.com/vincent-maillou/quatrex/pull/27](https://github.com/vincent-maillou/quatrex/pull/27)_

This PR was originally in quatrex repository (branch: Maurodossena/test qtbm).

Original PR description:

----

Added QTBM and some examples

---

## Tasks
- [x] Move away from run scripts; properly include QTBM in the CLI
- [x] k-point grid definition in the config instead of in the QTBM instantiation
- [x] Add an example for the carbon nanotube
- [ ] ~~Output options for orbital/atom resolved~~
- [x] Slight clean up of print statements and commented out code


almaeder:
is it possible to write a test for the injection vector? not sure what to test there

vetschn:
Some tasks i think should be done before merging:
- [ ] Move away from run scripts; properly include QTBM in the CLI
- [ ] k-point grid definition in the config instead of in the QTBM instantiation
- [ ] Add an example for the carbon nanotube
- [ ] Output options for orbital/atom resolved
- [ ] Slight clean up of print statements and commented out code

almaeder:
Before merging this PR, we should do:

- [x] Setting up an integration test.
- [ ] Small refactors to get a similar structure to the current SCBA.

tests/qttools/wave_function_solver/test_solve.py
```diff
@@ -0,0 +1,147 @@
...
+    a : sparse.csc_matrix
+        The generated sparse system matrix in the specified format.
+    b : NDArray
+        The right-hand side matrix with shape (n, m).
+
+    """
+    a = sparse.random(n, n, density=0.1, format=format, dtype=float)
+    a += 1j * sparse.random(n, n, density=0.1, format=format, dtype=float)
+    a += sparse.diags([2.0] * n, format=format, dtype=float)
+    b = xp.ones((n, m), dtype=xp.complex128, order=order)
```
almaeder, line 46: Why is the test RHS only ones?

src/quatrex/core/quatrex_config.py
```diff
@@ -362,13 +374,13 @@ class ElectronConfig(BaseModel):
     def set_left_right_fermi_levels(self) -> Self:
         """Sets the left and right Fermi levels if not already set."""
         if (self.left_fermi_level is None) != (self.right_fermi_level is None):
-            raise ValueError(
+            warnings.warn(
```
almaeder, line 377: I assume this was changed since QTBM uses the fermi levels from the contacts and doesnt set left/righ. Fermi levels should be refactored, but another quick fix would be to move this validator to QuatrexConfig and raise depending on the formalism. @vetschn what is your opinion?

src/quatrex/core/subsystem.py
```diff
@@ -92,7 +92,10 @@ def _configure_nevp(
                 use_pinned_memory=compute_config.nevp.use_pinned_memory,
             )
         if obc_config.nevp_solver == "full":
-            return Full(eig_compute_location=compute_config.nevp.eig_compute_location)
+            return Full(
+                eig_compute_location=compute_config.nevp.eig_compute_location,
+                reduce=compute_config.nevp.reduce_sparsity,
+            )
```
almaeder, line 98: Full in SCBA does currently not support the system size reduction. This will be added in a separate PR.

examples/carbon-nanotube-dist/quatrex_config.toml
```diff
@@ -1,18 +1,20 @@
 simulation_dir = "."
 
+formalism = "negf" # "negf" or "qtbm"
```
alexnick83: Isn't QTBM just NEGF without scattering? Shouldn't this be QTBM vs SCBA?

alexnick83: Do we actually need this variable? Couldn't we just check if there is an SCBA or QTBM entry in the configuration file?

examples/carbon-nanotube-qtbm/quatrex_config.toml
```diff
@@ -0,0 +1,60 @@
+simulation_dir = "."
+
+formalism = "qtbm" # "negf" or "qtbm"
+
+[device]
+transport_direction = "x"
+kpoint_grid = [1, 1, 1]
+num_orbitals_per_atom = { "X" = 1 }
```
alexnick83, line 8: Isn't this provided by the input data? Why do we need this?

src/qttools/kernels/inplace/cupy/__init__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2025 ETH Zurich and the authors of the qttools package.
+
+
+from qttools.kernels.inplace.cupy.inplace_add import (
+    add_kernel_comp,
```
alexnick83: Since all these kernels are operating in-place (I assume that is the reason for writing a custom implementation; cupy doesn't do in-place and uses extra memory), maybe call them `iadd` and `isub` to follow the same naming as the Python standard library?

src/qttools/kernels/inplace/cupy/inplace_add.py
```diff
@@ -0,0 +1,83 @@
...
+                            int i = blockDim.x * blockIdx.x + threadIdx.x;
+                            if (i < N) {
+                            M[ind[i]] += U[i];
+                            }
+                        }
+                        """,
+    "add_real_func",
+)
+
+add_OBC_inplace = cp.RawKernel(
```
alexnick83: Maybe interesting for @michael-wuethrich

src/qttools/kernels/linalg/eig.py
```diff
@@ -118,6 +134,144 @@ def _eig_cupy(
...
+
+@profiler.profile(level="debug")
+def _eig_nvmath_kernel(
+    A: NDArray,
+) -> tuple[NDArray, NDArray]:
+
+    # TODO: this will not work for real A
+    # and the cupy binding should be followed
+
+    # TODO: not recommended to create/destroy handle every time
```
alexnick83, line 197: Maybe make the handle a singleton and initialize it when you import cusolverDn.

src/qttools/kernels/linalg/eig.py
```diff
@@ -118,6 +134,144 @@ def _eig_cupy(
...
+    # Left eigenvectors (not computed)
+    vr = xp.zeros((n, n), dtype=xp.complex128, order="F")
+    info = xp.zeros(1, dtype=xp.int64)
+
+    yes_vector = nvmath.bindings.cusolver.EigMode.VECTOR
+    no_vector = nvmath.bindings.cusolver.EigMode.NOVECTOR
+
+    try:
+        # Query workspace size
+        lwork_device, lwork_host = cusolverDn.xgeev_buffer_size(
```
alexnick83, line 222: I am wondering whether the buffer-size calculation methods are slow enough that it would make sense to cache the result.

src/qttools/nevp/full.py
```diff
@@ -20,22 +22,95 @@ class Full(NEVP):
     ----------
     eig_compute_location : str, optional
         The location where to compute the eigenvalues and eigenvectors.
-        Can be either "numpy" or "cupy". Only relevant if cupy is used.
+        Can be either "numpy" or "cupy" or "nvmath".
     use_pinned_memory : bool, optional
         Whether to use pinnend memory if cupy is used.
         Default is `True`.
+    a_sparsity : tuple[NDArray, ...] | None, optional
```
alexnick83, line 29: This seems quite expensive. Why not use CSC format, which allows you to find zero columns with `indptr[1:] - indptr[:-1]`?

src/qttools/nevp/full.py
```diff
@@ -73,12 +148,29 @@ def __call__(self, a_xx: tuple[NDArray, ...]) -> tuple[NDArray, NDArray]:
         B = xp.kron(xp.tri(len(a_xx) - 2).T, xp.eye(a_xx[0].shape[-1]))
         A[:, : B.shape[0], : B.shape[1]] -= B
 
+        # Concatenate and delete
+        if self.reduce:
```
alexnick83, line 152: I think this needs some work. Is A sparse but with a different sparsity pattern than the NEVP coefficients?

src/qttools/obc/spectral.py
```diff
@@ -489,22 +476,28 @@ def __call__(
...
+                wrs_injected = ws[i, mask_injected_i]
+                dE_dk_injected = dEk_dk[i, mask_injected_i]
+
+                # Flux normalization
+                vrs_injected = vrs_injected / xp.sqrt(
+                    xp.real(dE_dk_injected[xp.newaxis, :])
+                )
+
+                # Compute surface phi
+                phi_surface.append(
```
alexnick83, line 498: Isn't the left-side matmul with `a_ji` (see line 504 on dev) missing here? Is that handled elsewhere now?

src/qttools/wave_function_solver/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2025 ETH Zurich and the authors of the qttools package.
```
alexnick83, line 1: @vincent-maillou think if there is some intersection here with @ohcomely 's work. For example, if we identify three contacts as a significant bottleneck, it would be interesting to do a study of the performance of different solvers in the actual QTBM workflow.

src/qttools/wave_function_solver/cudss.py
```diff
@@ -0,0 +1,234 @@
+# Copyright (c) 2025 ETH Zurich and the authors of the qttools package.
+
+import ctypes.util
+from collections.abc import Sequence
+
+from qttools import NDArray, sparse
+from qttools.profiling import Profiler
+from qttools.wave_function_solver.solver import WFSolver
+
+try:
```
alexnick83, line 10: nvmath is imported separately for eig and cuDSS. Consider adding a higher-level import similar to xp, since I expect we'll use it more and more.

src/qttools/wave_function_solver/cudss.py
```diff
@@ -0,0 +1,234 @@
...
+
+except ImportError:
+    nvmath_available = False
+
+
+profiler = Profiler()
+
+# Possible multithreading libraries for cuDSS in descending order of
+# preference.
+_mtlayer_libs = [
```
alexnick83, line 30: This looks like it will need a lot of maintenance. If nvmath is not handling it automatically, we should consider an abstraction for proper cuDSS installation (is it not part of the cudatoolkit?).

src/qttools/wave_function_solver/cudss.py
```diff
@@ -0,0 +1,234 @@
...
+    def __del__(self) -> None:
+        """Cleans up the cuDSS solver context."""
+        if self.direct_solver is not None:
+            self.direct_solver.free()
+            self.direct_solver = None
+
+    def _explicitely_reset_b(self, b: NDArray) -> None:
+        """Resets the right-hand side in the direct solver.
+
+        This is a workaround for the current limitation in nvmath where
```
alexnick83, line 95: Do we (always) need this functionality? Is there logic surrounding this method to avoid calling it when not necessary?

src/qttools/wave_function_solver/cudss.py
```diff
@@ -0,0 +1,234 @@
...
+                "Copying across memory spaces is not supported for resetting the right-hand side."
+            )
+        ds.b = b
+
+        # Update the rhs and result shape and strides.
+        ds.rhs_shape = shape
+        ds.rhs_strides = strides
+
+        ds.result_shape = shape
+        # NOTE: I do not know what these two lines do.
```
alexnick83, line 161: I read this as cuDSS supports a syntax for solving multiple linear systems with a single `A@x=b` solver statement, where A, x, and b are arrays of matrices/vectors. The rhs array may not necessarily have the same shape as the lhs; presumably, NumPy broadcasting rules apply.

src/qttools/wave_function_solver/mumps.py
```diff
@@ -0,0 +1,90 @@
...
+
+except ImportError:
+    mumps_available = False
+
+profiler = Profiler()
+
+valid_orderings = ["amd", "amf", "scotch", "pord", "metis", "qamd", "auto"]
+
+
+class MUMPS(WFSolver):
```
alexnick83, line 20: I would consider, in the future, discussing with @vincent-maillou and @ohcomely the potential integration with https://github.com/vincent-maillou/SolverBench (currently private), which will ultimately include installation scripts and bindings for solvers like MUMPS supporting Alps and some x86-based clusters. For now, just adding a few notes saying that this is experimental and please figure out yourselves how to install MUMPS (and METIS OR SCOTCH), and the python-mumps bindings suffice.

src/qttools/wave_function_solver/solver.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2024 ETH Zurich and the authors of the qttools package.
```
alexnick83, line 1: 2025, I guess. We should also start changing this to 2025-2026, but we can do a replacement and commit all files in another PR.

src/quatrex/cli/main.py
```diff
@@ -47,47 +47,63 @@ def version_callback(value: bool):
...
+def _run_qtbm(quatrex_config, compute_config):
+    """Runs quatrex with the given configuration.
 
     Parameters
     ----------
-    quatrex_config_path : Path
-        Path to the quatrex configuration file.
-    compute_config_path : Path | None, optional
-        Path to the compute configuration file, by default None.
+    quatrex_config : QuatrexConfig
```
alexnick83, line 55: I suggest you accept both a configuration object and a file (as previously). Or at the very least, let's sync on this because I also added something similar in the [better-batching PR](https://github.com/quatrex/quatrex/pull/172/changes/19f26dee511a7e7bedfd96ad6ac27382676d5948), needed for testing.

alexnick83: I see that you moved the file-handling to the CLI call, but not everything will go through there. Tests may (need to) use the `run` method directly. Of course, we could also have the tests handle the files or call some utility method; to discuss in more detail.

src/quatrex/cli/main.py
```diff
@@ -47,47 +47,63 @@ def version_callback(value: bool):
...
+        if comm.rank == 0:
+            typer.secho(f"Leaving QTBM after: {(toc - tic):.2f} s")
 
-    if compute_config_path is not None:
-        compute_config = parse_compute_config(compute_config_path)
-    else:
-        compute_config = ComputeConfig()
 
-    quatrex_config = parse_quatrex_config(quatrex_config_path)
+def _run_negf(quatrex_config, compute_config):
```
alexnick83, line 80: This is SCBA, right? Both are NEGF. Or let's clarify the naming.

src/quatrex/device/contact.py
```diff
@@ -0,0 +1,1199 @@
...
+from mpi4py.MPI import COMM_WORLD as comm
+
+from qttools import NDArray, obc, sparse, xp
+from qttools.nevp import NEVP, Beyn, Full
+from quatrex.core.compute_config import NEVPConfig
+from quatrex.core.kpoints import monkhorst_pack
+from quatrex.core.quatrex_config import OBCConfig
+
+
+class Contact:
```
alexnick83, line 15: Is this generalizable to all NEFG models? Probably TODO for another PR.

src/quatrex/device/contact.py
```diff
@@ -0,0 +1,1199 @@
...
+        coordinates = []
+
+        for y in range(-radius, radius + 1):
+            for z in range(-radius, radius + 1):
+                if max(abs(y), abs(z)) == radius:
+                    coordinates.append(np.array([y, z]))
+
+        return coordinates
+
+    def _init_hamiltonian_overlap_matrices(self) -> int:
```
alexnick83, line 495: Why is this here? How does it differ from the existing utilities for creating Hamiltonians from primitive unit cells?

src/quatrex/device/contact.py
```diff
@@ -0,0 +1,1199 @@
...
+
+import numpy as np
+from mpi4py.MPI import COMM_WORLD as comm
+
+from qttools import NDArray, obc, sparse, xp
+from qttools.nevp import NEVP, Beyn, Full
+from quatrex.core.compute_config import NEVPConfig
+from quatrex.core.kpoints import monkhorst_pack
+from quatrex.core.quatrex_config import OBCConfig
+
```
alexnick83, line 13: I feel that this file has way too many things that go quite beyond the main theme (a Contact class). I would leave a big refactor (that also tries to unify the QTBM- and SCBA-related utilities) for another PR, but we should at least split the utilities.

src/quatrex/device/device.py
```diff
@@ -0,0 +1,368 @@
...
+        self.apply_potential()
+        self._add_contacts()
+
+        if comm.rank == 0:
+            print(
+                f"Device initialized with {len(self.contacts)} contacts.",
+                flush=True,
+            )
+
+    def _init_hamiltonian(self) -> None:
```
alexnick83, line 166: This appears to load full Hamiltonians, and there is no option to construct them from primitive unit cells. However, primitive unit cells (UC) are mentioned in the contacts file, so I am bit confused on the input requirements of QTBM.

src/quatrex/grid/kpoints.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) 2026 ETH Zurich and the authors of the quatrex package.
```
alexnick83, line 1: Is this synced with the SCBA k-points PR?

src/quatrex/core/qtbm.py
```diff
@@ -0,0 +1,1123 @@
...
+from quatrex.core.statistics import fermi_dirac
+
+_preferred_matrix_type = {
+    "mumps": sparse.coo_matrix,
+    "superlu": sparse.csc_matrix,
+    "cudss": sparse.csr_matrix,
+}
+
+
+def kr_mat_mul(m: NDArray, a: NDArray, vect: NDArray) -> NDArray:
```
alexnick83: This probably belongs in qttools.

src/quatrex/core/qtbm.py
```diff
@@ -0,0 +1,1123 @@
...
+    -------
+    result : NDArray
+        Resulting vector from the multiplication.
+
+    """
+    vect_3d = vect.reshape(a.shape[0], m.shape[0], -1, order="F")
+
+    # 2. Apply 'a' to the first dimension (axis 0)
+    # tensordot(a, phi, axes=1) is like a @ phi along the first axis
+    temp = np.tensordot(a, vect_3d, axes=1)
```
alexnick83: Surely cupy has tensordot? Any reason this is happening on the CPU only?

src/quatrex/core/qtbm.py
```diff
@@ -0,0 +1,1123 @@
...
+    # 3. Apply 'm' to the second dimension (axis 1 of temp)
+    # We contract axis 1 of m with axis 1 of temp
+    res_simple = np.tensordot(temp, m, axes=(1, 1))
+
+    res_simple = res_simple.transpose(0, 2, 1).reshape(-1, vect.shape[1], order="F")
+
+    return res_simple
+
+
+def allocate_sys_mat(
```
alexnick83: This needs a GPU implementation. Maybe we should go through the logic.

src/quatrex/core/qtbm.py
```diff
@@ -0,0 +1,1123 @@
...
+    ham: dict, ovl: dict, boundary_SE_indexes: list[NDArray]
+) -> list[sparse.csr_matrix]:
```
alexnick83: Is the output a list? It doesn't seem so.

src/quatrex/core/qtbm.py
```diff
@@ -0,0 +1,1123 @@
...
+    )
+
+    # Check if SM has canonical format
+    if not system_matrix.has_canonical_format:
+        raise ValueError("System matrix is not in canonical format after allocation.")
+
+    return system_matrix
+
+
+def compute_update_indeces_sparse(
```
alexnick83: indices? (spelling)

src/quatrex/core/qtbm.py
```diff
@@ -0,0 +1,1123 @@
...
+            raise ValueError(
+                "Some destination indexes do not exist in the system matrix row, cannot compute update indices."
+            )
+
+        update_indices[row_start:row_end] = M_ind_map + M_row_start
+
+    return xp.array(update_indices)
+
+
+def compute_update_indeces_dense(
```
alexnick83: indices? (spelling)

src/quatrex/core/qtbm.py
```diff
@@ -0,0 +1,1123 @@
...
+    temp = np.tensordot(a, vect_3d, axes=1)
+
+    # 3. Apply 'm' to the second dimension (axis 1 of temp)
+    # We contract axis 1 of m with axis 1 of temp
+    res_simple = np.tensordot(temp, m, axes=(1, 1))
+
+    res_simple = res_simple.transpose(0, 2, 1).reshape(-1, vect.shape[1], order="F")
+
+    return res_simple
+
```
alexnick83: This file seems to be full of stuff that needs a good sparse implementation, especially on the GPU.

src/quatrex/core/qtbm.py
```diff
@@ -0,0 +1,1123 @@
...
+        comm.Barrier()
+        self.observables.electron_transmission_contacts = xp.concatenate(
+            comm.allgather(self.observables.electron_transmission_contacts), axis=-1
+        )
+        self.observables.electron_dos_orb = xp.concatenate(
+            comm.allgather(self.observables.electron_dos_orb), axis=-1
+        )
+
+        # Compute the current from all the k dependent transmissions
+        for n_t in range(self.num_transmissions):
```
alexnick83: Maybe this needs another compute-observables method?

alexnick83 (changes requested):
Overall, the code is well written. However, I see red flags in performance (to be addressed by working with students in subsequent PRs), and I am not convinced about the current hierarchy. At the very least, a lot of the methods should be moved to qttools as appropriate. We also need to discuss the overall workflow and how it aligns with SCBA (to be addressed in a subsequent PR). I would avoid any further development in this PR, aside from moving things around. It is already too big as-is.

examples/carbon-nanotube-dist/quatrex_config.toml
almaeder: Better naming would be "scba"/'qtbm" or "negf"/"wf".

examples/carbon-nanotube-qtbm/quatrex_config.toml
almaeder: The input data provides us with the hopping matrices (`hamiltonian_x_y_z.npz`),
but not for which points to solve.
Full hamiltonian in the system matrix would be ~`hamiltonian_0_0_0 + e^{iky}*hamiltonian_0_1_0 + e^{-iky}*hamiltonian_0_-1_0`. I believe k points in SCBA #205 are done in the same way.

We should also be able to directly input the full summed-up Hamiltonian, but this is for a subsequent PR.

src/qttools/kernels/linalg/eig.py
almaeder: We could also get the handle from CuPy since it should be interoperable.

src/quatrex/grid/kpoints.py
almaeder: Kinda, both are using Monkhorst pack to generate the kpoint grid.

tests/qttools/wave_function_solver/test_solve.py
vetschn: This should also work with random values i believe, so no real reason for this. Is one preferable over the other one?

examples/carbon-nanotube-dist/quatrex_config.toml
vetschn: I believe the naming logic was that in QTBM we never actually compute a Green's function, but in NE**GF** we do so, so the two are distinct from that point.

But I agree that checking whether there is a scba/qtbm entry in the config would likely be a better solution :+1:

src/qttools/wave_function_solver/cudss.py
vetschn: Unfortunately, at least when i was putting this together, it did not automatically detect an installed multithreading lib, and it ended up running single-threaded. The default seems to be to pass `multithreading_lib=None` into the `DirectSolver` instance (see docs [here](https://docs.nvidia.com/cuda/nvmath-python/0.5.0/host-apis/sparse/generated/nvmath.sparse.advanced.DirectSolverOptions.html#nvmath.sparse.advanced.DirectSolverOptions.multithreading_lib)).

cuDSS is not part of the cudatoolkit, it had to be installed seperately. I think depending on how `nvmath-python` is installed (conda/pip) it may have tried to also install it, but i would need to check again what the actual behavior is there

src/quatrex/core/quatrex_config.py
vetschn: >  Fermi levels should be refactored, but another quick fix would be to move this validator to QuatrexConfig and raise depending on the formalism.

I'm with you on this 👍 Fermi level should eventually be set per contact (so i guess they should end up in the `ContactConfig` for both SCBA and QTBM) and then validated depending on the transport formalism and the requested output (to compute Landauer-Büttiker current in QTBM you need the contacts' chemical potentials 💀 )

examples/carbon-nanotube-dist/quatrex_config.toml
almaeder: I would say checking if entry exist for scba/qtbm is also not great since it doesnt allow to nicely switch between different type of simulations. For example, I had a setup which manipulates programmatically the default config and dumps the result to a new config file which results in both scba and qtbm entries existing.

src/qttools/obc/spectral.py
almaeder: It was changed which quantities is  returned since qtbm needed both the injection vectors and the surface phi.
Now the injection vectors are outside compute instead of returning dependent quantities.

src/qttools/nevp/full.py
almaeder: It has a sparsity which can be determined from the coefficients:
image
(Sascha Brück et. al)

src/quatrex/device/contact.py
almaeder: Does currently not generalize due to the following:
- Simple contact description of left/right not possible or description through start/end index.
- Constructs OBC inputs without a possible having a self-energy.

src/quatrex/device/contact.py
almaeder: Difference is that the unit cells are "automatically"  discovered from the position and the hamiltonian elements.
There is no "unit_cells.npy" input which allows to have a periodic contact with an inhomogenous device,
but fully periodic device coming from "unit_cells.npy"  are  not  supported.

src/quatrex/device/contact.py
almaeder: Agree, the discovery of periodicity should be handled separately and
the contact should only know of the indices.
Then, the contact blocks are sliced during the `compute_boundary` call.

src/quatrex/device/device.py
almaeder: See https://github.com/quatrex/quatrex/pull/5#discussion_r2683981571.

src/quatrex/core/qtbm.py
almaeder: This could be a bug since the inputs seem to be created with `xp`.
Changed it to `xp`. @maurodossena could you verify on the GPU?

src/qttools/utils/inplace_utils.py
vetschn, line 1: Some docstrings missing in this file (e.g. `sub_inplace_OBC`)

src/qttools/wave_function_solver/cudss.py
vetschn: in QTBM we solve `m @ x = b`, the symbolic factorization of `m` stays the same for all energies (so `direct_solver.plan()`, which is the slow part on CPU, should be called only once per run). The number of injection vectors (which sets the number of columns in `b`) is determined during the OBC calculation and can be different from one energy to the next. That's the reason for doing this "explicit reset", allocating new `b` and `x` with different shapes.

An option completely avoiding this would be to somehow find an upper bound for the number of injection vectors over all energies. There could potentially be a way to determine this from the contact band structure, but that would need some work and validation.

A compromise could be to check if the number of injection vectors stays the same from one energy to the next. If it does, explicitly reset, and implicitly reset if it doesn't. We could even sort the order of energies after the OBC computation and before the solve, to group the number of injection vectors and minimize the explicit resets

src/qttools/wave_function_solver/cudss.py
vetschn: This makes a lot of sense 👍 Thanks!

examples/carbon-nanotube-dist/quatrex_config.toml
alexnick83: However, the presence of both SCBA and QTBM is erroneous (unless we want to switch between those models using the same config file, which sounds error-prone). @almaeder maybe check #162 where I am writing with tomli-w to do the same, but I have the option to automatically skip defaults (and also decide exatly which keys should be written).

examples/carbon-nanotube-qtbm/quatrex_config.toml
alexnick83: I mean the `num_orbitals_per_atom`. To rephrase, why didn't we need this dictionary in SBCA?

src/qttools/kernels/linalg/eig.py
alexnick83: True. We need to discuss how much we want to use nvmath. We likely want to use cupy when functionality is equivalent because we can hope that eventually there will be parity with rocm/hip.

src/qttools/wave_function_solver/cudss.py
alexnick83: Then we update this as cuDSS evolves (I imagine that they will eventually provide a more robust solution).

src/quatrex/device/contact.py
alexnick83: I see. If I understand correctly, this is necessary whenever you have a non-homogeneous device and need periodic contacts. I suppose this is independent of the model (i.e., we could run SCBA on such a device). In that case, I would try to move those to input utilities or whatever applies better.

examples/carbon-nanotube-qtbm/quatrex_config.toml
almaeder: Ah, it isn't currently needed in the SCBA since we have no notion of atoms.
The positions (`.xyz` file) and the potential are atome resolved qtbm, while the `grid.npy` is orbital resolved, I believe.
Instead, it should be possible to operate in a purely orbital picture, but some extra preprocessing would be needed, @maurodossena, right?

vetschn: The QTBM inputs are now very different from the SCBA (they need to be coerced together asap). There is an `.xyz` file that contains the atom positions in the device. In CP2K (and other codes) different atoms can have different numbers of orbitals depending on the basis set (so maybe 13 orbitals per carbon atom but another number for molybdenum). The elements of the corresponding Hamiltonian are for each orbital.  

The main reason to keep track of `num_orbitals_per_atom` is to then sum together charges (that we compute per orbital) into charge per atom, which will be needed for Poisson.

We have not needed this in SCBA yet, because we always dealt with Wannier functions and not atom-centered basis functions. We should also have this for SCBA.

src/qttools/kernels/linalg/eig.py
almaeder: This was included by @Kosheira1 since the compilation from source to get eig from cupy was problematic.
With newer versions of cupy, our own nvmath eig should be removed.

src/quatrex/device/contact.py
almaeder: Agree, but we should move the input utilities to another subfolder and separate different kinds.
Should we make the above move in this PR?
What I suggested should be better another one.

examples/carbon-nanotube-qtbm/quatrex_config.toml
maurodossena: > Ah, it isn't currently needed in the SCBA since we have no notion of atoms. The positions (`.xyz` file) and the potential are atome resolved qtbm, while the `grid.npy` is orbital resolved, I believe. Instead, it should be possible to operate in a purely orbital picture, but some extra preprocessing would be needed, @maurodossena, right?

Moving from an atoms picture to a pure orbital picture could be done with a very fast preprocessing step. I don't see any problem with that

src/quatrex/device/contact.py
maurodossena: I agree, the construction of contacts could be separated by the periodicity discovery

src/quatrex/core/qtbm.py
maurodossena: Good catch! I just copied this function from a prototype code and I forgot to convert `np` to `xp`. It should work also with cupy, I'll test that.

src/quatrex/core/qtbm.py
maurodossena: I'll be happy to discuss a possible GPU implementation!

src/qttools/kernels/linalg/eig.py
almaeder: Since the pre-release CuPy v14.0.0rc1 already contains EIG, I suggest removing the nvmath eig before merging. We shouldn't maintain our own binding.

src/quatrex/core/contact.py
```diff
@@ -0,0 +1,1091 @@
...
+                for k in range(self.number_of_transport_cells)
+            ],
+            dtype=int,
+        )[None, :]
+
+        self.obc_solver = self._configure_obc(
+            device.quatrex_config.electron.obc, device.compute_config.nevp
+        )
+
+        self._compute_band_structure()
```
almaeder: Now, every rank does the same contact band structure calculation.
The band structure is currently only used in `qtbm.py`, where it is saved to file.
I would suggest removing the contact band structure calculation for now and adding in a separate PR a "formalism" that only does this without having to do QTBM.

maurodossena: Agree! I added that just as a test for a specific calculation I was doing

src/qttools/obc/spectral.py
maurodossena, line 227: If I remember correctly, the function should support batchsize > 1. Maybe this ValueError comes from an older version of the function

almaeder: Agree, this could be due to the rebasing we done.

src/quatrex/core/contact.py
almaeder: I removed it.

src/qttools/kernels/inplace/cupy/__init__.py
alexnick83: @copilot can you identify those kernels and rename them appropriately? The renaming should be the following (but please validate):
- add_kernel_real/comp -> iadd_kernel_real/comp
- sub_kernel_real/comp -> isub_kernel_real/comp
- add_OBC_inplace -> iadd_OBC
- sub_OBC_inplace -> isub_OBC

alexnick83:
@copilot this PR introduces several new files under `src/quatrex/core`. We want to organize them a bit better. We would like to group the files `contact.py` and `device.py` under a `device` submodule. We also want to group `energies.py` and `kpoints.py` under a `grid` submodule. Can you move the files and adapt the code as appropriate?

examples/carbon-nanotube-qtbm/quatrex_config.toml
vetschn: Opened a new issue #214 where we will then work on this after merging the branch. Hope this is okay :)

src/qttools/kernels/linalg/eig.py
vetschn: I also prefer getting rid of our own custom wrapper for `cusolverDn.xgeev`. I don't expect that the one in the `cupy-14rc` has any meaningful difference in performance.

Since we check in `qtools.kernels.linalg.eig` whether `cupy` has `eig`, we already raise a `ValueError` in any case if cupy < v14, so no need to modify the `pyproject.toml`.

**Edit:**
Added a comment about this [here](https://github.com/quatrex/quatrex/issues/185#issuecomment-3748586395). We will keep it here until there is an official release of cupy 14.

src/qttools/nevp/full.py
vetschn: Concluding: If A is sparse, then modes corresponding to zero columns are zero/infinite (which we do not care about). The sparsity is different than the coefficients, but can be computed from them. (@almaeder approves this message)

src/qttools/obc/spectral.py
vetschn: We'll keep this restriction for now, as we have not tested the injection vector calculation with batchsize > 1 yet. (Adding a TODO in the code)

src/qttools/nevp/full.py
vetschn: See issue #215

src/qttools/wave_function_solver/cudss.py
vetschn: Considered, but for when we use more nvmath since eig is just temporary, but cudss is forever.

src/quatrex/cli/main.py
vetschn: I suggest we keep this as is for the time being. In the integration tests we indeed go through the `run` function whenever configs should be parsed.

src/quatrex/core/quatrex_config.py
vetschn: We'll deal with this when dealing with #187

alexnick83 (approved):
It looks good to me now, based on what we discussed as targets for completing the PR.

src/quatrex/device/contact.py
vetschn: See #187

src/quatrex/device/contact.py
vetschn: Maybe more cohesive to do this with #187

alexnick83 (approved):

PR #162, opened 2025-09-05
[vincent-maillou/qttools#158] Memory use improvements

almaeder:
**PR by [alexnick83](https://github.com/alexnick83)**
_Originally opened at 2025-06-25T13:14:34Z GMT_
_Originally opened as [https://github.com/vincent-maillou/qttools/pull/158](https://github.com/vincent-maillou/qttools/pull/158)_

This PR was originally in qttools repository (branch: mem-use-opt).

Original PR description:
Reduces unnecessary data copies when creating DSDBSparse arrays and introduces new utility methods.

This PR introduces a few memory use optimizations:
- [x] Uses new DSDBSparse API methods to defer data allocation or avoid extraneous data copies when generating from a sparse array.
- [x] More aggressively allocates and deallocates data during the SCBA loop.
- [x] Copies some data to the host when they are not needed and copies them back to the device when they are required again.

----

PR #172, opened 2025-09-05
Memory optimization and better batching of the OBC

### LINKED ISSUE #118: Memory use optimizations
<a href="https://github.com/alexnick83"></img></a> **Issue by [alexnick83](https://github.com/alexnick83)**
_Wednesday Jun 25, 2025 at 16:27 GMT_
_Originally opened as https://github.com/vincent-maillou/quatrex/pull/66_

----

This PR introduces a few memory use optimizations:
- [x] Uses new DSDBSparse API methods to defer data allocation or avoid extraneous data copies when generating from a sparse array.
- [x] More aggressively allocates and deallocates data during the SCBA loop.
- [x] Copies some data to the host when they are not needed and copies them back to the device when they are required again.

----

_**[alexnick83](https://github.com/alexnick83)** included the following code: https://github.com/vincent-maillou/quatrex/pull/66/commits_

almaeder:
- [x] Critical bug with the ~~electron~~ coulomb screening (P-related) observables
- [x] Saving the OBC on the host

Update 08.01.2026

The above issues have been resolved, albeit in a suboptimal way:
- P is not deleted anymore after finishing computing `VP` and `L`. The main issue here is that the P-related observables can be computed immediately after polarization, allowing early deletion of P. However, this requires changes to the API, so it is better to address it in a later PR.
- The electron OBCs now have a full buffer initialized (for all local energies). The electron OBCs are still computed using the new, improved batching flow, but they slice the full buffer appropriately. This is suboptimal because the buffer is kept on the device during GPU execution, which consumes more memory than strictly necessary. I opted not to implement another host-copy logic to first discuss whether a better solution exists, e.g., computing the electron observables per batch.


almaeder:
This PR superseeds #162:
**PR by [alexnick83](https://github.com/alexnick83)**
_Originally opened at 2025-06-25T13:14:34Z GMT_
_Originally opened as [https://github.com/vincent-maillou/qttools/pull/158](https://github.com/vincent-maillou/qttools/pull/158)_

This PR was originally in qttools repository (branch: mem-use-opt).

Original PR description:
Reduces unnecessary data copies when creating DSDBSparse arrays and introduces new utility methods.

This PR introduces a few memory use optimizations:
- [x] Uses new DSDBSparse API methods to defer data allocation or avoid extraneous data copies when generating from a sparse array.
- [x] More aggressively allocates and deallocates data during the SCBA loop.
- [x] Copies some data to the host when they are not needed and copies them back to the device when they are required again.


----

src/quatrex/coulomb_screening/polarization.py
```diff
@@ -126,6 +126,8 @@ def compute(
                     )  # 16 bytes for complex128
                     batch_size = max(min(batch_size, no), 1)
                     batches = int(np.ceil(no / batch_size))
+                    # QUICKFIX for persistent OOM in Polarization (or later) issue.
```
almaeder, line 129: Would it be better to further overapproximate the number of buffers?

src/qttools/comm/comm.py
```diff
@@ -552,6 +552,10 @@ def configure(
 
         self._is_configured = True
 
+    def is_configured(self) -> bool:
+        """Checks if the communicator is configured."""
+        return self._is_configured
+
```
vetschn, line 559: This would be a perfect fit for a `@property` decorator

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -886,7 +1041,28 @@ def zeros_like(cls, dsdbsparse: "DSDBSparse") -> "DSDBSparse":
...
+        return ndslice if isinstance(ndslice, tuple) else (ndslice,)
+
+    first = ensure_tuple(first)
+    second = ensure_tuple(second)
+
+    out = list(first) + [slice(None)] * (len(shape) - len(first))
+    remaining_dims = [i for i, s in enumerate(out) if isinstance(s, slice)]
+    for i, rhs in zip(remaining_dims, second):
+        out[i] = _compose_single(out[i], rhs, length=shape[i])
+    return tuple(out)
```
vetschn: Very neat! Some short docstrings would be great here, was not very clear what this is for at just a glance.

src/quatrex/core/scba.py
```diff
@@ -859,20 +943,66 @@ def run(self) -> None:
...
+        self.data.sigma_lesser.data[:] = 0.0
+        self.data.sigma_greater.data[:] = 0.0
+        self.data.sigma_retarded.data[:] = 0.0
+
         for i in range(self.quatrex_config.scba.max_iterations):
             print(f"Iteration {i}", flush=True) if comm.rank == 0 else None
             # append for iteration time
+            self.data.sigma_lesser._data[:] = 0.0
+            self.data.sigma_greater._data[:] = 0.0
+            self.data.sigma_retarded._data[:] = 0.0
```
vetschn: This will need to go 😄

src/quatrex/electron/sse_coulomb_screening.py
```diff
@@ -231,6 +231,8 @@ def compute(
                     )  # 16 bytes for complex128
                     batch_size = max(min(batch_size, no), 1)
                     batches = int(np.ceil(no / batch_size))
+                    # QUICKFIX for persistent OOM in SSE issue.
```
vetschn, line 234: Same point @almaeder raises in the polarization
> Would it be better to further overapproximate the number of buffers?

vetschn (changes requested):
Very nice! Will validate this again once the bug with the per-batch computation of the contact current is fixed 👍

almaeder:
There seems to be an issue with shapes when using full nevp solver for the W obc.
This should be investigated.

almaeder:
There also seems to be an issue with the batch size in the FFT. I am not sure if this is a general issue with the current code,
but OOMs happen while the FFT batch size is still large.

src/quatrex/coulomb_screening/polarization.py
alexnick83: I am not sure it helps much to keep increasing them. I think we should decrease them in practice by fusing the elementwise operations. I would leave this for another PR. I can also rollback the change, but I don't think it makes much difference.

src/quatrex/electron/sse_coulomb_screening.py
alexnick83: See response above.

alexnick83:
@almaeder please also comment again.

src/qttools/utils/gpu_utils.py
```diff
@@ -342,3 +385,25 @@ def free_mempool():
     if xp.__name__ == "cupy":
         mempool = xp.get_default_memory_pool()
         mempool.free_all_blocks()
+
+
+def debug_gpu_memory_usage(msg: str = "") -> None:
```
almaeder, line 390: How should we generally handle this? We could include it in the profiler #207, but currently, it will lead to many print statements. Maybe for now we could have an ENV variable to enable the memory debugging.

src/qttools/lyapunov/lyapunov.py
```diff
@@ -151,12 +153,12 @@ def _call_with_cache(
             The solution of the discrete-time Lyapunov equation.
 
         """
-        x = self.lyapunov_solver(a, q, contact, out=out)
+        x = self.lyapunov_solver(a, q, contact, stack_slice, out=out)
         if out is None:
-            self._cache[contact] = x.copy()
+            self._cache[(contact, stack_slice)] = x.copy()
```
almaeder, line 158: Wouldn't it be simpler to attach the stack slice to the contact name in the outermost level? Then no extra argument needs to be passed, which isn't used except in the caching logic. "contact" is this; we should then rename this argument.

src/quatrex/electron/solver.py
```diff
@@ -45,14 +54,14 @@ def _btd_subtract(a: DSDBSparse, b: DSDBSparse) -> None:
     b_ = b.stack[...]
     for i in range(a.num_local_blocks):
         j = i + 1
-        a_.blocks[i, i] -= b_.blocks[i, i]
+        a_.blocks[i, i] -= xp.asarray(b_.blocks[i, i])
```
almaeder, line 57: This is for when b is on the host?

src/quatrex/electron/solver.py
```diff
@@ -503,19 +565,33 @@ def _compute_obc(self) -> None:
...
+                self.obc_blocks.greater[0] = xp.empty_like(self.obc_blocks.retarded[0])
+
+            self.obc_blocks.retarded[0][stack_slice] = sigma_00
             gamma_00 = 1j * (sigma_00 - sigma_00.conj().swapaxes(-2, -1))
 
             # Compute and apply the lesser boundary self-energy.
-            self.obc_blocks.lesser[0] = 1j * scale_stack(
-                gamma_00.copy(), self.left_occupancies
+            self.obc_blocks.lesser[0][stack_slice] = 1j * scale_stack(
+                gamma_00.copy(), self.left_occupancies[stack_slice]
```
almaeder, line 590: I assume we keep here the lesser/greater obc alive to compute the observables?

almaeder: Should we by default, move the obc blocks to the host?
The assignment and host copy to the full obc blocks could be done after the system solve to only have a copy back to the device in `contact_currents`.

src/quatrex/core/scba.py
```diff
@@ -1026,6 +1160,9 @@ def run(self) -> None:
             # Update self-energy for next iteration with mixing factor.
             t_sigma_update_start = time.perf_counter()
             self._update_sigma()
+            self.data.sigma_lesser_prev = None
+            self.data.sigma_greater_prev = None
+            self.data.sigma_retarded_prev = None
```
almaeder, line 1165: Isn;t this happening in `_update_sigma` as well?

src/quatrex/core/scba.py
```diff
@@ -893,22 +1024,10 @@ def run(self) -> None:
...
-                    flush=True,
-                )
-                print(
-                    f"Time for swapping all: {t_stash_end_all - t_stash_start:.3f} s",
-                    flush=True,
-                )
+            if xp.__name__ != "cupy":
+                self.data.sigma_lesser_prev = self.data.sigma_lesser._data.copy()
+                self.data.sigma_greater_prev = self.data.sigma_greater._data.copy()
+                self.data.sigma_retarded_prev = self.data.sigma_retarded._data.copy()
```
almaeder, line 1030: Could the update `sigma_prev` logic be moved together for `np/cp` into the `_stash_sigma`?

almaeder (commented):
Great work to reduce the memory footprint!

almaeder:
The batching should be the same as not batching. It would make sense to change the carbon nanotube example to have batchsize smaller than 4 (`energy_window_num_per_rank = 4`) to capture the batching in the reference tests.
I would suggest a batchsize of 3.

src/qttools/utils/gpu_utils.py
alexnick83: A variable makes sense.

src/qttools/lyapunov/lyapunov.py
alexnick83: Ah I see, make a combined `contact-slice` input.

src/quatrex/electron/solver.py
alexnick83: Yes. I think I ended up using this to stream the Hamiltonian block-by-block to reduce memory consumption.

src/quatrex/electron/solver.py
alexnick83: So now, I reverted to a situation where `self.obc_blocks.xxx` has a full local-energy buffer. Therefore, we compute them in slices without discarding any data. For now, this is staying on the GPU. I am not copying to the CPU because I don't have a nice object-oriented mechanism yet (the rest is handled through the DSDBSparse class), and I think this PR is already large as-is. However, it should not be hard; just implement methods similar to those in DSDBSparse for the OBCBlocks class.

src/quatrex/coulomb_screening/polarization.py
almaeder: I just saw some issues in practice, but this should be part of another PR 👍.

src/quatrex/core/scba.py
alexnick83: It looks like it.

src/quatrex/core/scba.py
alexnick83: True

alexnick83:
> The batching should be the same as not batching. It would make sense to change the carbon nanotube example to have batchsize smaller than 4 (`energy_window_num_per_rank = 4`) to capture the batching in the reference tests. I would suggest a batchsize of 3.

There are already tests that do the cross product [1000, 2, 1]x[1000, 2, 1] for solver and RGF batches. However, maybe a test with a fixed total number of energies may be required to rreveal if there may be deadlocks due to improper use of the global communicator for certain barriers within batches (happened already in distributed RGF).

PR #186, opened 2025-10-07
Refactored memoization options

almaeder:
This PR gives slightly more control over the memoization.
It includes an additional option to directly force by using either $q$ or $(M_{00})^{-1}$ 
as a starting guess for Lyapunov or OBC.

### DIFF
--- a/examples/carbon-nanotube-dist/quatrex_config.toml
+++ b/examples/carbon-nanotube-dist/quatrex_config.toml
@@ -63,8 +63,7 @@ obc.r_i = 0.8
 obc.m_0 = 31
 obc.num_quad_points = 31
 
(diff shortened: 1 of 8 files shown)


vetschn (approved):
Beyn test fails again in the CI... This looks good though! Thanks

almaeder:
We should replace the beyn test with one that uses the CNT data. 
But this is something for the next PR.

vetschn:
> We should replace the beyn test with one that uses the CNT data.

Yeah, that may be a good alternative. I think we should generally reconsider whether randomized test data is the best way to check all the numerical methods, or at least seed everything consistently.

maurodossena (approved):
Looks good!

PR #188, opened 2025-10-08
Qtbm devnv

Kosheira1:
Pull request that would allow the use of the nvmath binding for xgeev for the full NEVP boundary solver.
Reason for this change is that it wasn't possible for me to create a consistent environment that installs cupy from source (requires CDK 13) and does a conda install of nvmath (requires CDK<13 to install libcudss directly).

I left the eig_compute_location as cupy because you use this variable to sometimes mean to target GPU memory and sometimes to refer to the cupy module so the call to _eig_nvmath looks a bit unclean. I let you decide how you want to handle it. Müsst Ihr wissen.


Kosheira1:
sorry i try to rebase

PR #191, opened 2025-10-20
CLI Changes

vetschn:
Some clean up of the CLI.


src/quatrex/cli/main.py
```diff
@@ -83,33 +91,64 @@ def run_quatrex(
             print(f"Leaving SCBA after: {(toc - tic):.2f} s")
 
 
-@quatrex_cli.command("fetch-example")
+@quatrex_cli.command()
+def run(
+    quatrex_config: Annotated[
```
almaeder, line 95: Should we allow that `quatrex run` with no inputs, is okay, and it auto-detects the files?

almaeder (commented):
LGTM

almaeder:
Tests fail since this is a problem with HuggingFace

src/quatrex/cli/main.py
vetschn: This is now fixed. It will search for both `quatrex_config.toml` and `compute_config.toml` in the working directory if no arguments are provided. If the only argument provided to `quatrex run` is a directory (e.g. `quatrex run .`), both files are looked for in that directory.

almaeder: Thank you

almaeder (approved):
LGTM

PR #192, opened 2025-10-21
Change treat_pairwise default value to False

maurodossena:
This pull request makes a small configuration change to the `OBCConfig` class in `src/quatrex/core/quatrex_config.py`. The default value for the `treat_pairwise` option has been changed from `True` to `False`.

`treat_pairwise` makes the OBC calculation very unstable for some devices (metal contacts, large nanowires).


almaeder (approved):
🔥

PR #193, opened 2025-10-23
Change inv kernel calls

maurodossena:
Cupy inversion kernel might be slow in case of batching when a big block size is present. By switching to our custom inv kernel (`linalg.inv`), disbatching is automatically performed when needed (it calls the cupy `solve` function).


vetschn (approved):
I found one last occurance of an `xp.linalg.inv()` call in the "full" NEVP solver. @maurodossena maybe you can check again if Beyn is still faster than full with this change (i think you mentioned this, right?)

almaeder (approved):
Straightforward improvement

PR #196, opened 2025-10-27
Refactor obc tests to use real example data

almaeder:
This refactored the obc tests to use data from the CNT example.
The example is fetched in the conftest and contact blocks are extracted.

Additionally, tests of non-recommended features are removed.

### DIFF
--- a/tests/qttools/obc/conftest.py
+++ b/tests/qttools/obc/conftest.py
@@ -3,21 +3,20 @@
 import pytest
 
 from qttools import NDArray, xp
-from qttools.datastructures.dsdbsparse import _block_view
 from qttools.nevp import NEVP, Beyn, Full
+from qttools.utils.gpu_utils import get_host
+from qttools.utils.mpi_utils import distributed_load
+from quatrex.cli.main import fetch_example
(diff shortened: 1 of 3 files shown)


tests/qttools/obc/conftest.py
```diff
@@ -78,45 +71,54 @@ def batch_size(request: pytest.FixtureRequest) -> int:
...
-@pytest.fixture(params=TREAT_PAIRWISE)
-def treat_pairwise(request: pytest.FixtureRequest) -> bool:
-    """Whether the eigenvalues are pairwise filtered."""
-    return request.param
+    hamiltonian_sparray = distributed_load(
+        quatrex_config.input_dir / "hamiltonian.npz"
+    ).astype(xp.complex128)
+    block_sizes = get_host(
+        distributed_load(quatrex_config.input_dir / "block_sizes.npy")
+    )
```
vetschn: Not 100% sure if this is the best idea to reload the data at every call of the fixture, maybe check if there is a better way

almaeder: `scope="session"` should fix this

vetschn (approved):

PR #197, opened 2025-10-28
Refactored NEVP tests

almaeder:
Refactored the nevp tests such that the same data as in the obc tests is used.
Additionally, testing for the left eigenvectors is removed, which will be deprecated.


vetschn (commented):
Looks good but the folder structure seems weird to me. Now you seem to have this:
```
nevp_obc_system
|- conftest.py
|- nevp
|  |- test_nevp.py
|- obc
   |- conftest.py
   |- test_sancho_rubio.py
   |- test_spectral.py
```
Could you not do this instead?
```
obc
|- conftest.py
|- test_sancho_rubio.py
|- test_spectral.py
|- nevp
   |- test_nevp.py
```
Is there a problem if the nevp tests see same conftest as the obc?

(You could also just remove the nevp folder completely, no?)

vetschn (approved):
👍

PR #198, opened 2025-10-28
Removing non recommended features

almaeder:
This PR removes features from the spectral obc/nevp computations that are not recommended
and thus should not be used:

- two_sided / left keyword: only using the right eigenvectors for the pseudo inverse seems stable and is less expensive.
- out keyword: There is currently no use case for pre-allocating memory for the OBC. Memory has to be allocated inside anyways.
- only the "self-energy" formula: Sascha Brück described two formulas for the surface Green's function. Experimentally determined, the "self-energy" formula is more stable/correct.
- treat_pairwise keyword: pairwise treating the modes to determine which are the reflected ones is unstable. Additionally, this can not work when the bands are not symmetric. 
- Max decay is now determined in the configuration if not set and not inside the spectral obc solver.
- The residual normalization with the operator norm is removed, as it is too expensive.


maurodossena (approved):
Looks good! Nice cleanup!

src/qttools/obc/spectral.py
```diff
@@ -273,7 +180,6 @@ def _find_reflected_modes(
         ws: NDArray,
         vrs: NDArray,
```
vetschn: Small note: Rename `vrs` to `vs`

vetschn (commented):
Very good, simplifies things a lot. Just a small note on the v**r**s.

vetschn (approved):
LGTM :shipit:

PR #202, opened 2025-11-04
Improved filtering for ballistic simulations

almaeder:
This add an extra condition to capture slow decaying modes in the obc filtering.
This helps to reduce the recursion error in the obc in the ballistic case.


src/qttools/obc/spectral.py
maurodossena, line 296: I would add the following condition also (to match the condition for propagating modes):

```
mask_injected &= self.min_propagation < abs(dEk_dk.real) / (
            abs(dEk_dk.imag) + eta
        )
```

maurodossena (changes requested):

maurodossena (approved):
Looks good!

PR #205, opened 2025-11-11
Kpoints rebased

### LINKED ISSUE #98: Inclusion of kpoints
<a href="https://github.com/awinka"></img></a> **Issue by [awinka](https://github.com/awinka)**
_Tuesday Dec 17, 2024 at 11:24 GMT_
_Originally opened as https://github.com/vincent-maillou/quatrex/pull/18_

----

First attempt at including k-points. An example with MoS2 is included.

----

_**[awinka](https://github.com/awinka)** included the following code: https://github.com/vincent-maillou/quatrex/pull/18/commits_

awinka:
Rebased version of the kpoints to the newest dev branch. It contains the inclusion of k-points, new features enabling this, and a set of bug fixes.

Bug fixes:
 - Re-computes the diagonal indices when changing block sizes (currently only for DSDBCOO)
 - Checks that the sparsity pattern is symmetric when performing symmetrization

New features:
 - In-place addition and subtraction of the _DStackView (needed for the assembly of the k-point dependent Hamiltonian and Coulomb matrix).
 - Added option for performing sandwhich calculation in case the `a` matrix isn't symmetric (used in calculation of $W^{\lessgtr}=W^r P^{\lessgtr} W^a$ in bandstructure calculation)
 - Computation of "current conservation" condition.
 - Separated update of sigma with memory factor and symmetrization of sigma.
 - Hilbert transform for polarization
 - Added functions for loading the input matrices, also the Coulomb matrix is loaded in the scba script, instead of doing it both for the screened interaction and fock self-energy.
 - Contructs a `tanh` potential profile when there is a applied bias and no potential profile is provided (this should maybe change)

Left to do:
 - [x] Check symmetrization
 - [x] Should probably remove the "adiabatic" turn on of the Coulomb matrix, did not see any benefit for it.
 - [x] Move FFTs to qttools
 - [x] Add tests


src/qttools/convolutions/ffts.py
```diff
@@ -0,0 +1,223 @@
...
+    b : NDArray
+        Second array.
+
+    Returns
+    -------
+    NDArray
+        The convolution of the two arrays.
+    """
+    ne = a.shape[0] + b.shape[0] # Should not have -1 here (otherwise hilbert transform fails)
+    a_fft = xp.fft.fftn(a, (ne,), axes=(0,))
```
alexnick83: Is this the same as calling `xp.fft.fft(a, ne, axis=0)`?

alexnick83: NOTE: We should check if there is a performance difference between calling 1D FFT and ND FFT with 1 pseudo-time dimension and the rest batches.

src/qttools/fft/ffts.py
```diff
@@ -0,0 +1,223 @@
...
+    a_fft = xp.fft.fftn(a, nka, axes=axes)
+    b_fft = xp.fft.fftn(b, nkb, axes=axes)
+    return xp.fft.ifftn(a_fft * b_fft, axes=axes)
+
+
+@profiler.profile(level="api")
+def fft_convolve_kpoints(a: xp.ndarray, b: xp.ndarray) -> xp.ndarray:
+    """Computes the convolution of two arrays using the FFT.
+
+    The first axis is assumed to be the energy axis, the other
```
alexnick83, line 66: Are there any particular considerations in deciding the order? For example, why not first the k-points, then the energy, and last the orbitals?

src/qttools/convolutions/ffts.py
```diff
@@ -0,0 +1,223 @@
...
+        xp.fft.ifftn(a_fft * b_fft, axes=(0,) + tuple(range(1, len(nka) + 1))),
+        shift=1,
+        axis=tuple(range(1, len(nka) + 1)),
+    )
+
+
+@profiler.profile(level="api")
+def hilbert_transform_polarization(a: NDArray, energies: NDArray) -> NDArray:
+    """Computes the Hilbert transform of the array a, assuming the symmetries of the
+    polarization, i.e $[P^{\lessgtr}_{ij}(\omega)]^{\dagger} = -P^{\gtrless}_{ij}(-\omega)$.
```
alexnick83: Will LaTeX math render in the documentation, or do we need special syntax and plugins? @vetschn do you know?

src/qttools/convolutions/ffts.py
```diff
@@ -0,0 +1,223 @@
...
+
+    hilbert_kernel = 1 / energy_differences
+    b = fft_convolve(a, hilbert_kernel)[:ne]
+    # Negative frequencies of a
+    b += fft_convolve(a[::-1].conj(), hilbert_kernel)[-ne:]
+    # Negative frequencies of the kernel
+    hilbert_kernel = -hilbert_kernel[::-1]
+    b += fft_convolve(a, hilbert_kernel)[-ne:]
+
+    #hilbert_kernel_fft = xp.fft.fft(
```
alexnick83: Is this equivalent to the logic above?

src/qttools/convolutions/ffts.py
```diff
@@ -0,0 +1,223 @@
...
+    sr = fft_convolve(sg[:ne] - sl[-ne:], hilbert_kernel)[:ne]
+    # Correct for left edge
+    sr += fft_convolve(-sl[:ne], hilbert_kernel)[-ne:]
+    # Next account for negative frequencies
+    hilbert_kernel = -hilbert_kernel[::-1]
+    sr += fft_convolve(sg[:ne] - sl[-ne:], hilbert_kernel)[-ne:]
+    # Correct for right edge
+    sr += fft_convolve(sg[-ne:], hilbert_kernel)[:ne]
+
+    return sr
```
alexnick83:
```suggestion
    return sr

```

src/qttools/datastructures/dsdbcoo.py
```diff
@@ -750,6 +760,26 @@ def spy(self) -> tuple[NDArray, NDArray]:
             cols[i] += rank_offset[i]
         return xp.hstack(rows), xp.hstack(cols)
 
+    def _check_sparsity_pattern_symmetric(self) -> bool:
```
alexnick83, line 768: The operation, implemented as-is, will be slow. Aside from any efforts to improve it, wouldn't it make more sense to do this check only once per input sparsity pattern (read: not every sparsity pattern needs to be checked, i.e., those that result from the composition of sparsity patterns should be OK) rather than invoking it every time we want to symmetrize?

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -983,6 +1003,48 @@ def __setitem__(self, index: tuple[ArrayLike, ArrayLike], values: NDArray) -> No
         rows, cols = self._dsdbsparse._normalize_index(index)
         self._dsdbsparse._set_items(self._stack_index, rows, cols, values)
 
+    def __iadd__(self, other: "DSDBSparse | sparse.spmatrix") -> "DSDBSparse":
```
alexnick83, line 1006: Not blocking, but is it possible that we need to operate among StackViews? In #172, we do need to compose them, so I am wondering if we will end up performing arithmetic operations among them, too. Let's discuss whether we need to make a revised design for these classes.

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -983,6 +1003,48 @@ def __setitem__(self, index: tuple[ArrayLike, ArrayLike], values: NDArray) -> No
...
+        if sparse.issparse(other):
+            csr = other.tocsr()
+            self._dsdbsparse.data[self._stack_index] += xp.squeeze(
+                xp.asarray(csr[self._dsdbsparse.spy()])
+            )
+            return self._dsdbsparse
+        try:
+            # TODO: Lots more checks should be done here.
+            # For example, the nnz sizes should match.
+            self._dsdbsparse.data[self._stack_index] += xp.squeeze(
```
alexnick83, line 1017: What does `squeeze` do here? I guess it is an issue when the stack index is a single element; other.data has a energy dimension of size 1 and you get dimension mismatch unless you squeeze?

src/qttools/datastructures/routines.py
```diff
@@ -261,9 +266,16 @@ def bd_sandwich(
                         a_k, a_j = k, j
                 if ab_ik[k] is None:
                     continue
-                partsum += (ab_ik[k] @ a_.blocks[a_k, a_j]).astype(
-                    accumulator_dtype
-                )  # cast data type
+                if symmetric:
```
alexnick83: Is this a bug fix, i.e., did the symmetry-update PR break the non-symmetric case?

alexnick83: If this is the case, wouldn't it make more sense to consult the symmetry attributes in DSDBSparse rather than having another parameter in the product method?

src/quatrex/bandstructure/band_edges.py
```diff
@@ -247,7 +247,12 @@ def find_renormalized_eigenvalues(
             rank_left = xp.digitize(ind_left, section_offsets) - 1
 
             if rank_left == comm.stack.rank:
-                local_ind = ind_left - section_offsets[rank_left]
+                # NOTE: This assumes that each rank has all k-points and that the band edge
```
alexnick83, line 250: I see now why you use {energy, k-points, orbitals} order. However, I don't think this is a reasonable assumption (for example, imagine you want to perform k-point calculations on the large Si nanowires). Let's discuss solutions and performance considerations.

src/quatrex/core/observables.py
```diff
@@ -72,21 +74,29 @@ def density(x: DSDBSparse, overlap: sparse.spmatrix | None = None) -> NDArray:
 
     """
     if overlap is None:
-        local_density = x.diagonal().imag
+        local_density = x.diagonal()  # .imag
```
alexnick83: Is this a bug fix or a generalization for k-points?

src/quatrex/core/quatrex_config.py
```diff
@@ -315,6 +315,8 @@ class ElectronConfig(BaseModel):
 
     model_config = ConfigDict(extra="forbid")
 
+    number_of_kpoints: tuple[PositiveInt, PositiveInt, PositiveInt] = (1, 1, 1)
```
alexnick83: Maybe this should be `num_kpoints` or something like that to be consistent with the naming of other variables?

src/quatrex/core/quatrex_config.py
```diff
@@ -487,6 +490,15 @@ class DeviceConfig(BaseModel):
     number_of_supercells: PositiveInt = 1
     transport_direction: Literal["x", "y", "z"]
 
+    R_cutoff: tuple[int, int, int] | None = None
+    # Periodicity of the unit cell in non-transport directions.
```
alexnick83: @almaeder is this the extra periodicity that is also added in #5?

src/quatrex/core/scba.py
```diff
@@ -117,13 +128,23 @@ def __init__(
...
+        if max_interaction_cutoff > 0.0:
+            self.sparsity_pattern = compute_sparsity_pattern(
+                grid,
+                max_interaction_cutoff,
+                transport_direction=quatrex_config.device.transport_direction,
+                start_idx=start_idx,
+                end_idx=end_idx,
+            )
+        else:
+            # If there is no interaction cutoff (no scattering), we use the identity matrix.
```
alexnick83: Maybe this can be absorbed into `compute_sparsity_pattern`.

alexnick83: Since we have a few standard (overlap) or fallback (this) identity matrices, perhaps it would be worth making a PR in the future that adds special handling for those instead of using a generic sparse matrix. @vetschn what do you think?

src/quatrex/core/scba.py
```diff
@@ -364,6 +388,53 @@ def __init__(
 
         # ----- Coulomb screening --------------------------------------
         if self.quatrex_config.scba.coulomb_screening:
+            # Load the Coulomb matrix.
```
alexnick83, line 342: I guess we should also discuss this soon. I think it is OK for now to push the PR, but it may be too limiting to require doing all k-points on every rank.

src/quatrex/core/scba.py
```diff
@@ -475,10 +558,27 @@ def _stash_sigma(self) -> None:
...
+    @profiler.profile(level="api")
+    def _symmetrize_sigma(self) -> None:
+        # Symmetrization.
+        if not self.quatrex_config.scba.symmetric:
+            self.data.sigma_lesser.symmetrize(xp.subtract)
+            self.data.sigma_greater.symmetrize(xp.subtract)
+            # Make the self-energy Hermitian (removing the skew-Hermitian part).
+            self.data.sigma_retarded.symmetrize(xp.add)
+
+        # self.data.sigma_lesser._data.real = 0
```
alexnick83: Is the zeroing now done somewhere else, is it wrong to do it, or is it a different approach? If the last, perhaps we should gather together all these different approaches (for example, the P-retarded approach from @jiangcao) and discuss possible special-case handling.

src/quatrex/core/scba.py
```diff
@@ -532,9 +607,18 @@ def _has_converged(self) -> bool:
         dE = self.electron_energies[1] - self.electron_energies[0]
         current_diff = xp.abs(xp.sum(i_left) * dE + xp.sum(i_right) * dE)
 
+        current_conservation_abs = self.observables.electron_current.get(
```
alexnick83: Maybe this should be guarded here. Are we always computing (do we want to?) the current conservation?

src/quatrex/core/scba.py
```diff
@@ -939,6 +1024,14 @@ def run(self) -> None:
                     flush=True,
                 )
 
+            if self.quatrex_config.scba.coulomb_screening:
+                if (
+                    i >= 1
+                    and i < self.quatrex_config.coulomb_screening.num_adiabatic_steps
+                ):
+                    self.coulomb_screening_solver.coulomb_matrix.data *= (i + 1) / i
```
alexnick83: I am wondering whether two copies of the Coulomb matrix start to become expensive.

src/quatrex/core/utils.py
```diff
@@ -26,6 +31,253 @@ def homogenize(matrix: DSDBSparse) -> None:
...
+        lattice_matrices_array = xp.array([lattice_matrix[cell] for cell in valid_cells])
+    # Check if it is sparse.csr_matrix or sparse.coo_matrix
+    elif isinstance(lattice_matrix[valid_cells[0]], sparse.csr_matrix):
+        lattice_matrices_array = np.array([lattice_matrix[cell] for cell in valid_cells])
+    elif isinstance(lattice_matrix[valid_cells[0]], sparse.coo_matrix):
+        lattice_matrices_array = np.array([lattice_matrix[cell].tocsr() for cell in valid_cells])
+    else:
+        raise TypeError("Unsupported lattice matrix type.")
+    
+    # Loop over k-points
```
alexnick83: Perhaps we can use itertools (or a self-made iterator) to abstract away an N-dimensional for-loop.

src/quatrex/coulomb_screening/solver.py
```diff
@@ -135,32 +103,46 @@ def __init__(
...
+        dummy_dsbsparse = compute_config.dsdbsparse_type.from_sparray(
+            sparsity_pattern.astype(xp.float32),
+            block_sizes=self.small_block_sizes,
+            global_stack_shape=(comm.size,),
+            symmetry=quatrex_config.scba.symmetric,
+            symmetry_op=xp.conj,
+        )
         v_times_p_sparsity_pattern = _compute_sparsity_pattern(
-            self.coulomb_matrix, self.coulomb_matrix, dtype=xp.float32
+            dummy_dsbsparse, dummy_dsbsparse, dtype=xp.float32
```
alexnick83, line 117: Doesn't the coulomb matrix have the same sparsity pattern for some default k-value? Maybe we should make these methods work with a StackView (if they don't already).

src/quatrex/electron/solver.py
```diff
@@ -88,47 +94,16 @@ def __init__(
...
-                distributed_load(quatrex_config.input_dir / "block_sizes.npy")
-            )
+        hamiltonian_sparray, hamiltonian_dict, self.block_sizes = (
+            load_matrix_from_unit_cell(quatrex_config, "hamiltonian")
+            if quatrex_config.device.construct_from_unit_cell
+            else load_matrix_from_files(quatrex_config, "hamiltonian")
+        )
+        # Symmetrize the Hamiltonian sparse array. (Noticed that sparsity pattern
+        # was not always symmetric).
+        hamiltonian_sparray = 0.5 * (hamiltonian_sparray + hamiltonian_sparray.conj().T)
```
alexnick83: That's quite expensive in memory. This should be done only for matrices loaded from a file. Do we expect issues with unit cells? If yes, I would symmetrize the unit cells instead and then construct the matrix.

src/quatrex/electron/solver.py
```diff
@@ -315,6 +340,31 @@ def __init__(
             quatrex_config.electron.filtering_iteration_limit
         )
 
+    @staticmethod
```
alexnick83, line 222: I have the impression we already have multiple copies of this. Maybe we need to place it in some general sparse utilities.

src/quatrex/electron/solver.py
```diff
@@ -450,7 +508,8 @@ def _update_fermi_levels(
             left_conduction_band_edge - self.delta_fermi_level_conduction_band
         )
         self.right_fermi_level = (
-            right_conduction_band_edge - self.delta_fermi_level_conduction_band
+            self.left_fermi_level - self.bias
```
alexnick83: Is this equivalent, a bug fix, or a different approach?

src/quatrex/electron/solver.py
```diff
@@ -575,15 +634,15 @@ def _assemble_system_matrix(self, sse_retarded: DSDBSparse) -> None:
...
         self.system_matrix += self.overlap_sparray
         scale_stack(
             self.system_matrix.data,
             self.local_energies + 1j * self.eta,
         )
-
         self.system_matrix -= sparse.diags(self.potential, format="csr")
-        _btd_subtract(self.system_matrix, sse_retarded)
         _btd_subtract(self.system_matrix, self.hamiltonian)
+        _btd_subtract(self.system_matrix, sse_retarded)
```
alexnick83, line 426: Does the order matter?

src/quatrex/electron/sse_coulomb_screening.py
```diff
@@ -256,96 +185,347 @@ def compute(
...
-                    # imaginary part in the real domain, we have one less fft to do
-                    # antihermitian_fft *= self.prefactor * 0.5
-                    # antihermitian_fft -=  xp.flip(antihermitian_fft.conj(), axis=0) # remove the hermitian part X(-t) = X(t).conj()
-
-                    sigma_x_fft = xp.multiply(antihermitian_fft, hilbert_kernel_fft)
-                    sigma_x_fft -= xp.multiply(
-                        antihermitian_fft, hilbert_kernel_fft.conj()
-                    )  # negative energy part
+                    # Lesser self-energy
+                    sl = self.prefactor * fft_correlate_kpoints(
```
alexnick83: This looks clean, but it also computes the FFTs multiple times. To discuss

tests/qttools/convolutions/conftest.py
```diff
@@ -0,0 +1,13 @@
...
+
+ARRAY_SHAPES = [
+    pytest.param((20,), id="1D-array"),
+    pytest.param((20, 5), id="2D-array"),
+    pytest.param((20, 5, 3), id="3D-array"),
+]
+
+@pytest.fixture(params=ARRAY_SHAPES)
+def array_shape(request) -> tuple[int, ...]:
+    return request.param
```
alexnick83:
```suggestion
    return request.param

```

tests/qttools/convolutions/test_convolutions.py
```diff
@@ -0,0 +1,302 @@
...
+    @pytest.mark.usefixtures("array_shape")
+    def test_hilbert_transform_polarization(self, array_shape):
+        # Use the symmetry of the polarization $a(-\omega)=a^{*}(\omega)$
+        ne = array_shape[0]
+        # Add empty orbital dimension at the end
+        a = xp.random.random(array_shape+(1,)) + 1j * xp.random.random(array_shape+(1,))
+        energies = xp.linspace(-10, 10, ne)
+        result = hilbert_transform_polarization(a, energies)
+        expected = naive_hilbert_transform_polarization(a, energies)
+        assert xp.allclose(result, expected)
```
alexnick83:
```suggestion
        assert xp.allclose(result, expected)

```

alexnick83 (changes requested):
Overall, it is very well written. I have two main comments:
- There is the assumption that a rank holds for some energies all k-points. I think this is fine for this PR, but it will be a significant limitation going forward, so we should discuss subsequent work to address it.
- I see potential performance regressions for the non-kpoint version. We should discuss if we want to leave them for another PR or handle them in this one.

src/qttools/convolutions/ffts.py
```diff
@@ -0,0 +1,223 @@
...
+    #    2 * ne,
+    #    axis=0,
+    #)
+    #a_t = xp.fft.fft(a, 2 * ne, axis=0)
+    #b = xp.fft.ifft(
+    #    a_t * hilbert_kernel_fft
+    #    - a_t * hilbert_kernel_fft.conj()
+    #    - a_t.conj() * hilbert_kernel_fft,
+    #    axis=0,
+    #)[:ne]
```
almaeder:
```suggestion
```

src/qttools/convolutions/ffts.py
```diff
@@ -0,0 +1,223 @@
...
+    return b# / (2 * xp.pi) * (energies[1] - energies[0])
```
almaeder:
```suggestion
    return b
```

src/qttools/datastructures/dsdbcoo.py
```diff
@@ -656,6 +664,7 @@ def block_sizes(self, block_sizes: NDArray) -> None:
                 ]
                 + self.global_block_offset
             )
+            self._set_diagonal_indices()
```
almaeder, line 672: Would it make sense to make a caching logic to not recompute the indices?

src/qttools/datastructures/routines.py
```diff
@@ -150,8 +150,9 @@ def bd_sandwich(
     spillover_correction: bool = False,
     accumulator_dtype=None,
     accumulate: bool = False,
+    symmetric: bool = True,
 ):
-    """Compute the sandwich product `a @ b @ a` BTD DSDBSparse matrices.
+    """Compute the sandwich product `a @ b @ a.dagger()` BTD DSDBSparse matrices.
```
almaeder, line 154: Agree `a @ b @ a.dagger()` and not `a @ b @ a` should be performed.

almaeder: In the current use case `L<> = V @ P<> @ V^H`, it matters less since `V` is hermitian?

src/qttools/utils/mpi_utils.py
```diff
@@ -91,7 +92,7 @@ def distributed_load(path: Path) -> sparse.spmatrix | NDArray:
     """
     if not path.exists():
         raise FileNotFoundError(f"File not found: {path}")
-    if path.suffix not in [".npz", ".npy"]:
+    if path.suffix not in [".npz", ".npy", ".pkl"]:
```
almaeder: Which data will we read as pickles?

src/quatrex/core/quatrex_config.py
almaeder: I believe this is more similar to the `hamiltonian_x_y_z.npy`, because it is used to get a dict of the hamiltonians and then in the `assemble_kpoint_dsb`.
The extra periodicity #5 is for when "inside" a  `hamiltonian_x_y_z.npy`, the contact is periodic.

almaeder: This can then be used as an approximation to subslice the OBC into multiple, but smaller problems.  The approximation is correct if the full device is periodic in that dimension.

src/quatrex/coulomb_screening/polarization.py
```diff
@@ -203,12 +203,14 @@ def compute(
...
             p_lesser.symmetrize(xp.subtract)
             p_greater.symmetrize(xp.subtract)
+            p_retarded.symmetrize(xp.add)
 
         # Discard the real part.
-        p_lesser.data.real = 0
-        p_greater.data.real = 0
+        # p_lesser.data.real = 0
+        # p_greater.data.real = 0
+        # p_retarded._data[:] = 0
```
almaeder: This should break the current reference tests

src/quatrex/electron/solver.py
```diff
@@ -239,24 +228,60 @@ def __init__(
...
+                - quatrex_config.electron.left_fermi_level
+            )
+            # Map grid transport to the range [-1, 1]
+            grid_transport = (
+                2
+                * (grid_transport - grid_transport[0])
+                / (grid_transport[-1] - grid_transport[0])
+            ) - 1
+            # Create a potential profile
+            self.potential = bias / 2 * (xp.tanh(3 * grid_transport) + 1)
```
almaeder: @vetschn I thought it was intended to have control over Fermi levels separate from the potential if a poison solver is not present.

almaeder (commented):
Great work

src/qttools/datastructures/dsdbcoo.py
alexnick83: I suppose we can do that as well, in the same caching mechanism we use for block-size changes.

src/qttools/convolutions/ffts.py
awinka: Yes, it should be the same. It is more suitable to use `xp.fft.fft` here since it is only over one dimension

src/qttools/fft/ffts.py
awinka: This is just how the `DSDBSparse.data` is structured, so no particular reason

src/qttools/convolutions/ffts.py
awinka: This was how I implemented the hilbert transform for P before, but it doesn't give the same as the calculation above (which I believe is the correct one)

src/qttools/datastructures/routines.py
awinka: Yes, correct

src/qttools/datastructures/routines.py
awinka: Not a bug fix, in quatrex the coulomb matrix is symmetric adding this doesn't change anything. I also use this function in the `awinka/bandstructure` branch to compute the lesser/greater screened interaction with W^R P^{<>} [W^R]^{\dagger}, where W^R doesn't have any symmetry.

src/qttools/utils/mpi_utils.py
awinka: I also had data when the R-dependent Hamiltonian is sparse instead of dense (npy), so I used pickle to save it as a dict of sparse matrices. Maybe there is a better way to do this?

src/quatrex/core/observables.py
awinka: This was mostly for debugging purposes were I also wanted to look at the real part of the retarded quantities.

src/quatrex/coulomb_screening/solver.py
awinka: Yes, you could probably use the coulomb matrix here, unless floating numbers cause problems?

alexnick83: ~~It shouldn't be an issue, it should only be taking the rows and cols, but I will verify.~~ It actually computes the product. So there can be two issues, one with the dtype (fixable) and one with whether you need to pass a specific k-point and whether the StackView interface has all the necessary methods. It should work in better-batching.

src/quatrex/core/utils.py
awinka: Yes, that would be nice. This function is a bit messy atm

src/qttools/datastructures/dsdbcoo.py
awinka: I'll have a look at how it is done for the block sizes and try something similar for the diagonal indices

src/qttools/datastructures/dsdbcoo.py
awinka: Yes, that would be nice. Like an instance attribute?

src/qttools/datastructures/dsdbsparse.py
awinka: Yes, I think that the right-hand-side had an extra dimension of size 1, but I don't fully remember

src/quatrex/electron/solver.py
awinka: The unit cells are symmetric. I think the problem comes from floating point operations, the elements that break the symmetry are < 1e-16

src/quatrex/electron/solver.py
awinka: It is a different approach. It allows you to have a fermi level difference in case there is no drop in the potential profile (in which case the right and left conduction band edges would be the same)

src/quatrex/electron/solver.py
awinka: No, I think I swapped them when debugging. I will switch back

src/quatrex/core/scba.py
awinka: More a different approach. Physically they can be non-zero and shouldn't be zeroed, but from my experience it can help with stability to force them to zero. We should probably examine this further

src/quatrex/core/scba.py
awinka: This is mostly for reassurance that the self-energy is computed correctly, so it can be guarded

src/quatrex/core/scba.py
awinka: Yes, a single instance of the coulomb matrix would be nice. Now both the screening solver and fock self-energy have a copy

src/qttools/convolutions/ffts.py
vetschn: Math is rendered with MathJax in the docs but i'm not positive that it'll render with the `$...$` delimiters. For sure it works with `\(...\)` / `\[...\]`

**Edit:**
Works with both!

src/qttools/utils/mpi_utils.py
vetschn: Linking #214, we can discuss possible formats there :)

src/qttools/convolutions/ffts.py
```diff
@@ -0,0 +1,213 @@
...
+    ne = (
+        a.shape[0] + b.shape[0]
+    )  # Should not have -1 here (otherwise hilbert transform fails)
+    a_fft = xp.fft.fft(a, ne, axis=0)
+    b_fft = xp.fft.fft(b, ne, axis=0)
+    return xp.fft.ifft(a_fft * b_fft, axis=0)
+
+
+@profiler.profile(level="api")
+def fft_circular_convolve(a: xp.ndarray, b: xp.ndarray, axes: tuple[int]) -> xp.ndarray:
```
vetschn: Type hints are inconsistent here. As an aside we should probably replace the current `NDArray` definition via `TypeAlias` with the PEP 695 type parameter syntax, like

```python
type NDArray[ScalarType: xp.generic] = xp.ndarray[Any, xp.dtype[ScalarType]]
```
or even something like numpy does
```python
type NDArray[ScalarType: xp.generic] = xp.ndarray[tuple[Any, ...], xp.dtype[ScalarType]]
```

src/quatrex/core/quatrex_config.py
```diff
@@ -572,6 +599,15 @@ class DeviceConfig(BaseModel):
...
     kpoint_shift: tuple[float, float, float] = (0.0, 0.0, 0.0)
 
+    R_cutoff: tuple[int, int, int] | None = None
+    # Periodicity of the unit cell in non-transport directions.
+    # Only in one direction (so actuall number of periodic cells is 2 times minus 1 this value).
+    cells_in_periodic_directions: tuple[PositiveInt, PositiveInt, PositiveInt] = (
+        1,
+        1,
+        1,
+    )
```
vetschn: I think these parameters and their combination with the `unit_cell_per_supercell` setting open up some strange configuration options.

Is what we really want not just the cutoff radius for the considered image cells (so what is now `R_cutoff`)? The `unit_cell_per_supercell` follows from this (it should be the minimal number of cells that then lead to a BTD tiling of the system matrix) and the `cells_in_periodic_directions` are also just the ones that still include connections to image cells in transverse direction up to the given cutoff.

The only thing that becomes a bit difficult then is controlling the number of transport cells, but this should anyway be inferred through a setting like `device_length` (in Angstrom) and we just approximate that.

src/quatrex/core/quatrex_config.py
```diff
@@ -341,6 +341,8 @@ class ElectronConfig(BaseModel):
 
     model_config = ConfigDict(extra="forbid")
 
+    number_of_kpoints: tuple[PositiveInt, PositiveInt, PositiveInt] = (1, 1, 1)
```
vetschn: We should reuse the `kpoint_grid` from the `DeviceConfig` here. We will always have the same number of kpoints for electrons and screened Coulomb interaction, right? Same thing for other particles (or no k-points at all in the case of photons) as far as i can tell, correct? Otherwise we would need to do some strange interpolation

src/quatrex/core/quatrex_config.py
awinka: Yes, currently it is confusing .

`cells_in_periodic_directions` can be removed, and maybe also `unit_cell_per_supercell`. One reason for maybe keeping `unit_cell_per_supercell` is if we want a transport cell that is longer in the transverse directions than a single unit cell . For example in OMEN only neighboring transport cells are taken into account in transverse directions, so if you wanted to include all interactions the transport cell had to be "thicker". But I think this is something that we can drop

src/quatrex/core/scba.py
vetschn: I agree that we should handle the case of no interactions in a seperate PR. Besides this there are surely other things where special handling for identity overlap matrices is necessary :+1:.

I'm also not sure that just diagonal sparsity here is the right way to handle this. With this sparsity pattern we will not be able to get the (non-Meir-Wingreen) device current, for instance.

awinka: Having a zero interaction cut-off could give a local interaction (so identity matrix or overlap matrix). You can still calculate a device current, no?

src/qttools/utils/input_utils.py
vetschn: These functions are never used in the code, except for the tests. At the moment, we never load Wannier data in directly, it's expected that this is done as a preprocessing step. Would it be okay to remove these functions for now, and add them back in down the line? Maybe this can be handled in #214, when we define the input format more clearly.

awinka: Yes, sure. I use them to construct the input matrices, but they can be removed for now

src/qttools/datastructures/dsdbcoo.py
almaeder: See #223.

src/qttools/datastructures/dsdbcoo.py
vetschn: Rewrote this, and only do the check in the init.

src/qttools/datastructures/dsdbsparse.py
almaeder: Agree. We opened a discussion at #224.

src/qttools/fft/convolve.py
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2024-2026 ETH Zurich and the authors of the quatrex package.
+
+from qttools import NDArray, xp
+
+
+def _naive_convolve(a: NDArray, b: NDArray) -> NDArray:
```
vetschn, line 6: This seems to only be used in the tests and it does something not exactly the same as `np.convolve` with `mode` `"full"`, but almost the same (output shape is one off(?)). Why do our fftconvolve functions use this convention that is different from numpy?

It looks like it has to do with the convention that is used for the Hilbert transform. Revisit this at some point, maybe when implementing other physics.

src/quatrex/core/quatrex_config.py
```diff
@@ -572,13 +626,46 @@ class DeviceConfig(BaseModel):
     kpoint_grid: tuple[PositiveInt, PositiveInt, PositiveInt] = (1, 1, 1)
     kpoint_shift: tuple[float, float, float] = (0.0, 0.0, 0.0)
 
+    orthogonal_basis: bool = True
```
almaeder, line 629: This should also be used in the case of QTBM. Overlap is supported for QTBM, but less for SCBA.

src/quatrex/coulomb_screening/polarization.py
```diff
@@ -181,7 +225,7 @@ def compute(
         t_all2all2_start = time.perf_counter()
         # Transpose the matrices to stack distribution.
         with profiler.profile_range("nnz->stack transpose", level="debug"):
-            for m in (p_lesser, p_greater):
+            for m in (p_lesser, p_greater, p_retarded):
```
almaeder: This doesnt need to be done in the case of `self.compute_retarded=False`.

src/qttools/fft/convolve.py
awinka: Our `fft_convolve` is only used in the Hilbert transform, where the extra element in the output acts as a shift of one. We compute the Hilbert transform for first positive energies, and then for negative energies, and to have the correct result when adding these two contributions on top of each other they need to be shifted by a single element relative to each other (now that I think about it, we should probably do something similar when computing the self-energy). 

To make this shift more apparent, it should probably be handled by the Hilbert transform function instead of having this weird output shape

src/quatrex/core/scba.py
almaeder: We opened an issue for this. See #225.

src/quatrex/electron/solver.py
vetschn: See #226

almaeder (approved):
LGTM

vetschn (approved):
Great stuff :shipit:

alexnick83 (approved):

PR #207, opened 2025-12-04
Refactor the profiler to replace the timing prints

almaeder:
This PR refactors the old profiler:

- Used the assumptions of having only a single GPU to simplify the code
- Changed the profiling/profiling_range to have the same behavior as the prints (sync/time/sync/time/barrier/time)
- Replaces the prints with the profiler
- Dump the profiling results at the end of the CLI run
- Write the prints in the profiling to a separate file to clean up the outputs
- Simplify the resulting stats that are dumped

This PR should be merged only after GW-kpoints / better batching is done, since it will cause many conflicts.


almaeder:
@vetschn @alexnick83 the timing prints are currently appended if the output file already exists.
Are you okay with this behaviour, or should every `quatrex run` overwrite the timing print file?

examples/w90/carbon-nanotube/gw-dist/quatrex_config.toml
vetschn, line 1: Did you change some parameters here? We really should stop tracking this with lfs...

src/qttools/greens_function_solver/rgf_dist.py
```diff
@@ -270,179 +291,94 @@ def selected_solve(
...
                 xl_buffer_upper=xl_buffer_upper,
                 xl_out=xl_out_,
                 # Greater quantities.
-                sigma_greater=sigma_greater_,
                 xg_diag_blocks=xg_diag_blocks,
-                # xg_buffer_lower=xg_buffer_lower,
+                xg_buffer_lower=xg_buffer_lower,
                 xg_buffer_upper=xg_buffer_upper,
                 xg_out=xg_out_,
-                selected_solve=True,
```
vetschn, line 438: Diff is just cooked around here, or are there actual changes?

src/qttools/profiling/profiler.py
```diff
@@ -197,7 +222,17 @@ def __new__(cls):
             cls._instance = super(Profiler, cls).__new__(cls)
 
             cls._instance.eventlog = []
-            cls._instance.devices = get_cuda_devices()
+            cls._instance.depth = -1
+            cls._instance.print_file = _OutputFile("qtx_times")
+            cls._instance.save_path = "qtx_times"
```
vetschn: `quatrex_times` better? Everywhere else it is spelled out (like `quatrex_config.toml` etc.)

src/qttools/profiling/profiler.py
```diff
@@ -363,60 +349,82 @@ def profile(self, level: str = QTX_PROFILE_LEVEL):
...
-                    self._synchronize_events(start_events)
+                if xp.__name__ == "cupy":
+                    if NVTX_AVAILABLE:
+                        xp.cuda.nvtx.RangePush(label)
 
-                    # Record start events for each device.
-                    self._record_events(start_events)
+                    self.start_event.record(xp.cuda.get_current_stream())
+                    self.start_event.synchronize()
+                    self.start_event.record(xp.cuda.get_current_stream())
```
vetschn: Yes, looks way better to assume only one device 👍

src/qttools/profiling/profiler.py
```diff
@@ -12,36 +12,44 @@
...
+            self.file_handle = open(new_name, "a")
+            self.is_custom_file = True
+        except Exception:
+            self.file_handle = sys.stdout
+            self.is_custom_file = False
+
+    def write(self, message):
+        print(message, flush=True, file=self.file_handle)
+
+    def __delete__(self):
```
vetschn: Are you sure you want `__delete__` here and not `__del__`? Could you quickly explain what this is really supposed to do?

vetschn (changes requested):
Seems very nice! Only thing to me would be changing the name of the default output file to something like `quatrex_timings` or just `timings`. Also no file extension for this?

examples/w90/carbon-nanotube/gw-dist/quatrex_config.toml
almaeder: `profiling_stats = false` was changed to `save_profiling_results = true`

src/qttools/greens_function_solver/rgf_dist.py
almaeder: I believe the diff is cooked.

examples/w90/carbon-nanotube/gw-dist/quatrex_config.toml
almaeder: We should exclude text files from LFS.

src/qttools/profiling/profiler.py
almaeder: Gemini :shaking-fist:. It should be `__del__`.

vetschn (approved):
Looks great! I guess all that's left to do is make the linter happy

vetschn (approved):

Issue #178, opened 2025-09-16
More examples and symlinks to share data


almaeder:

vetschn:
Other things concerning examples:
- [ ] Switch to "point-charge" Coulomb matrix for the CNT and set an appropriate epsilon in the config.
- [ ] Make the fetching of example data a bit cleaner (naming and file hierarchy in the manifest.yaml)

In addition i still think it's a bit clunky to have the "compute" and "quatrex" configs separately (i think just a single config file would be nicer, this would also remove the need for the corresponding CLI options).

The CLI should also tell the user what datasets are available, and we should document the example workflow in the online docs (see #179)

almaeder:
[#217](https://github.com/quatrex/quatrex/pull/217) closed this.

PR #217, opened 2026-01-23
Refactoring of examples and reference test suite.

vetschn:
The way the examples and the reference tests were previously handled was quite annoying.

We (@almaeder and I) tried to improve things a bit:

- Now all example inputs, outputs, and configuration files are tracking using `git lfs`. We do not use GitHub's own LFS  service, but rather use the one from ETH's GitLab instance through this project: https://gitlab.ethz.ch/quatrex/quatrex. There we do not (seem to) have any bandwidth caps that would limit the number of CI test runs.
- The whole `fetch-example` functionality is removed, as it was a bit half-baked from the get-go. Since everything is tracked with LFS the integration tests can now just run the examples and compare the generated outputs with the reference outputs.
- We discovered some bugs here and there that we fixed as well. Notably, the resolving of directories in the config was a problem and we also changed the way random numbers are seeded for Beyn (instance-level rather than module-level).

Edit:
- We still have a memory storage limitation of 10GB, but this isn't an issue since our tests are in the ~1MB range.
- Beyn RNG is changed to have the same result if one calls the `run` API multiple times. This manifested in the `test_single_rank` test, where, due to noise, the results differed more than the tolerance allowed.
- This also removes the previous problem of having to update reference test data when the formulation changes manually/separately.
We also reviewed other methods, like git submodules or other manual tracking ideas, but they did not fit our case.


almaeder:
We should add an explanation to the documentation about installing git lfs before cloning.

almaeder (commented):
Great works!

almaeder:
This will close https://github.com/quatrex/quatrex/issues/178.

alexnick83 (approved):
Looks good to me

PR #227, opened 2026-02-19
Bug fix: Profiler paths were relative and not resolved

almaeder:
There was a bug introduced in #207:
- The paths for the profiler were not resolved.


vetschn (approved):
Looks like it could work now! 👍

PR #236, opened 2026-02-24
Do not track .toml files with LFS

almaeder:
Previously, the config files were tracked with LFS. 
This does not allow tracking the changes.
For other text files, it does not matter, but it does for the configs since they are often updated.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 examples/**/* filter=lfs diff=lfs merge=lfs -text
+
+examples/**/*.toml text filter= diff= merge=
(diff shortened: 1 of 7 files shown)


.gitattributes
```diff
@@ -1 +1,3 @@
 examples/**/* filter=lfs diff=lfs merge=lfs -text
+
+examples/**/*.toml text filter= diff= merge=
```
vetschn:
```suggestion
examples/**/* filter=lfs diff=lfs merge=lfs -text
examples/**/*.toml text filter= diff= merge=

```

Missing newline

vetschn (changes requested):

vetschn (approved):
Thanks! Small change, big improvement

PR #238, opened 2026-03-03
fixes indentation error in sse, should be part of top level if block

### LINKED ISSUE #237: Scope-related error in SSE
https://github.com/quatrex/quatrex/blob/13c15205e42f8177be45b7881a9aab0c423a4dc3/src/quatrex/electron/sse_coulomb_screening.py#L311-L312
...
https://github.com/quatrex/quatrex/blob/13c15205e42f8177be45b7881a9aab0c423a4dc3/src/quatrex/electron/sse_coulomb_screening.py#L347-L353

The scope starting at line 351 must be within the scope starting at line 312. Otherwise, if there are no ij elements to work on, the code will crash at line 352, because `batch_displacements` has not been set (line 347).

YongDaLi:
fix #237 indentation error, should be part of top level if block

### DIFF
--- a/src/quatrex/electron/sse_coulomb_screening.py
+++ b/src/quatrex/electron/sse_coulomb_screening.py
@@ -348,41 +348,41 @@ class SigmaCoulombScreening(ScatteringSelfEnergy):
                     np.concatenate(([0], np.array(batch_counts)))
                 )
 
-            if self.apply_hilbert_correction:
-                for start, end in zip(batch_displacements, batch_displacements[1:]):
-                    self._compute_with_correction(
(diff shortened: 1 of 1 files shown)


YongDaLi:
Reference from `polarization.py`, which follows the same structure with the top level if block.

https://github.com/quatrex/quatrex/blob/13c15205e42f8177be45b7881a9aab0c423a4dc3/src/quatrex/coulomb_screening/polarization.py#L134-L200

almaeder (approved):
LGTM

almaeder:
Thank you for the bug fix :fire:

PR #239, opened 2026-03-05
Bug in spectral if blocks are zero

almaeder:
- Potentially, the data is not symmetrical zero.
- Then the resulting blocks could have an even length and the eigenvalue problem is not well defined.
- This fixes this.

### DIFF
--- a/src/qttools/obc/spectral.py
+++ b/src/qttools/obc/spectral.py
@@ -2,6 +2,8 @@
 
 import warnings
 
+import numpy as np
+
 from qttools import NDArray, xp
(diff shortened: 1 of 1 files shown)


vetschn (approved):
Looks good to me. Does one of the test cases actually cover this? Might be nice to have this explicitely covered.

almaeder:
> Looks good to me. Does one of the test cases actually cover this? Might be nice to have this explicitly covered.

No test covers this explicitly. Also, I need to check if the code works on the GPU due to `xp.any`.

Issue #206, opened 2025-11-12
Test with mpiexec


almaeder:
Testing the command line interface with mpiexec is not great
when the tests are run on a Slurm-based system.

PR #240, opened 2026-03-05
Fix gpu profiler bug

almaeder:
- The profiler is used nested which leads to that the events are overwritten.
- This was fixed by doing a deviceSync and only host timing.
- This should be good enough since the profiler is only used coarse grained.
- Thus, the events can also be removed.


vetschn (approved):
Great! The timings now look more meaningful, or rather they actually make sense now.

Running full test suite on alps, i see all tests passing except for a few unrelated ones invoking `mpiexec` (see #206) and the `test_fft_circular_convolve` due to the use of `xp.random.choice` (see #242 ).

Issue #214, opened 2026-01-20
Consolidate different inputs for NEGF and WF simulations


vetschn:
Now we have to completely different input data formats:
- For WF we have a number of Hamiltonian/overlap files in a tight-binding picture (`hamiltonian_x_y_z.npy`) and the atom/orbital sites are given in (extended) xyz format.
- For NEGF we only give the "central" Hamiltonian along with grid in binary format and the block-sizes of the matrix.

We need to bring both input decks into a common format to allow users to more easily switch between the two transport formalisms.

This is somewhat related to #187 but the two should not be handled on the same feature branch.

### Tasks
- [x] Decide on the actual format to go with
- [x] Change parsing logic accordingly
- [x] Adjust the example inputs

vetschn:
We can handle the tight-binding inputs in two ways: upscaling small "unit cell" arrays internally or reading full, sparse device Hamiltonians that are already unfurled along the transport direction.

For the structure, a dense 5D array `(a, b, c, orb, orb)` has been fine for the small cells, but it can potentially be wasteful for some structures since you can end up storing a lot of zeros. A dictionary mapping cell indices to dense and sparse matrices would be more efficient.

So for storage, after looking at several options, I think [`scipy.io.savemat`](https://docs.scipy.org/doc/scipy-1.17.0/reference/generated/scipy.io.savemat.html) could be the best fit. It handles dictionaries of sparse (and dense) matrices natively, avoids the security/compatibility headaches of pickling, and keeps the data accessible in a pretty standard format. One caveat is that the dictionary keys have to be strings, so there has to be a little bit of conversion after loading the matrices.

This is how a loaded file could potentially look:
```python
>>> from scipy.io import loadmat
>>> h_r = loadmat("h_r.mat")
>>> h_r
{'__header__': b'MATLAB 5.0 MAT-file Platform: posix, Created on: Mon Jan 26 17:11:23 2026',
 '__version__': '1.0',
 '__globals__': [],
 '[-1 -1 -1]': <COOrdinate sparse matrix of dtype 'complex128'
 	with 300000 stored elements and shape (1000, 1000)>,
 '[-1 -1  0]': <COOrdinate sparse matrix of dtype 'complex128'
 	with 300000 stored elements and shape (1000, 1000)>,
 '[-1 -1  1]': <COOrdinate sparse matrix of dtype 'complex128'
 	with 300000 stored elements and shape (1000, 1000)>,
...

```

[...]

PR #241, opened 2026-03-05
Input bugs

almaeder:
There were multiple bugs which led to the code not working with gamma only and on the GPU.
These bugs are adressed in this PR:
- Not only check `index = np.argwhere(kpoint_grid > 1)[0]`, but checking for all ones.
- Replacing `xp` with `np` if needed.
- Removing `xp.conj` on a sparse matrix.
- Adding a gamma only example.

### DIFF
--- a/src/quatrex/device/inputs.py
+++ b/src/quatrex/device/inputs.py
@@ -413,7 +413,8 @@ def _assemble_kpoint(
             stack_index = tuple(stack_index[index])
 
             cells = np.array(list(matrix_dict.keys()))
-            phases = xp.exp(2j * xp.pi * (cells @ kpoint))
+            phases = np.exp(2j * np.pi * (cells @ kpoint))
+            phases = xp.asarray(phases)
 
             # NOTE: Sparse matrix addition is slow
             # but unavoidable due to memory constraints.
--- a/src/quatrex/device/inputs.py
+++ b/src/quatrex/device/inputs.py
@@ -386,7 +386,7 @@ def _assemble_kpoint(
     num_dimensions = len(kpoint_grid)
 
     if isinstance(kshift, int):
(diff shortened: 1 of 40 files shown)


src/quatrex/device/inputs.py
```diff
@@ -574,15 +574,18 @@ def load_matrix(
         sparsity_pattern = sparsity_pattern + sparsity_pattern.T
 
     # Symmetrize the data.
-    matrix_sparray = 0.5 * (matrix_sparray + symmetry_op(matrix_sparray).T)
+    # TODO: This should be avoided due to the extra copy
+    # when addressing issue #214, only the upper part should be kept
+    # as only symmetric matrices are loaded
+    matrix_sparray = 0.5 * (matrix_sparray + matrix_sparray.T.conj())
```
vetschn, line 580: Yeah, #214 will have to touch this again 👍

src/quatrex/device/inputs.py
```diff
@@ -525,10 +528,9 @@ def load_matrix(
     matrix_name: str,
     sparsity_pattern: sparse.coo_matrix | None = None,
     shift_kpoints: bool = False,
-    symmetry_op: Callable = xp.conj,
 ) -> tuple[DSDBSparse, sparse.coo_matrix]:
-    """Loads a matrix from file, applying symmetrization and optionally
-    using a provided sparsity pattern.
+    """Loads a hermitain matrix from file and optionally
+    apply a provided sparsity pattern.
```
vetschn:
```suggestion
    """Loads a Hermitian matrix from file and optionally
    applies a provided sparsity pattern.
```

vetschn (approved):
Great stuff, also nice addition of a unit cell example. #214 will lead to some adjustments here again but this is perfect for now.

Tested on alps, everything looks alright.

PR #244, opened 2026-03-06
Change cupyx.jit import and usage in operator.py

vetschn:
Import resolution changed in new version of cupy, making things break here on GPU. This should work both with old and new version.

In the other files implementing `jit.rawkernel`s we were already doing `from cupyx import jit` so no changes there.


almaeder (approved):
Thank you for the fix :rocket:

PR #245, opened 2026-03-07
Consolidate inputs

almaeder:
This PR addresses the issue #214.
The inputs for both SCBA and QTBM are consolidated by:
- Using `.mat` files
- Switching to `.xyz` for the lattice and the grid.


vetschn (approved):
Tested things on alps, and fixed some small GPU bugs. Looks good to me but would be great if someone else could have a look over this since i changed some stuff too.

PR #249, opened 2026-03-10
Try to improve the sparsity reduction in the full NEVP solver

### LINKED ISSUE #215: Improve sparsity reduction in full eigenvalue solver
> This seems quite expensive. Why not use CSC format, which allows you to find zero columns with `indptr[1:] - indptr[:-1]`? 

 _Originally posted by @alexnick83 in [#5](https://github.com/quatrex/quatrex/pull/5/changes#r2681801056)_

Currently, a tuple of dense mask arrays (`int`) has to be passed to the `Full` NEVP solver. This should be replaced with a tuple of sparse matrices.

This 'sparsity reduction' functionality is also not being used in SCBA, so the parameter should be used there as well.

vetschn:
Now we pass in tuples of sparse matrices, from which the sparsity pattern of the matrix A is inferred.

If the coefficient block sparsity patterns cannot be given at instantiation (this is the case in how SCBA is currently coded) the reduced sparsity can be computed at the first call.

Adjusted tests and exposed the functionality in SCBA and QTBM.

### DIFF
--- a/src/qttools/nevp/full.py
+++ b/src/qttools/nevp/full.py
@@ -3,6 +3,7 @@
 import warnings
 
 from qttools import NDArray, sparse, xp
+from qttools.comm import comm
 from qttools.kernels import linalg
 from qttools.nevp.nevp import NEVP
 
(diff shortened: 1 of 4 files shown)


src/qttools/nevp/full.py
almaeder: We should call it `inds` everywhere. I believe that is the current naming convention.

src/qttools/nevp/full.py
```diff
@@ -21,92 +21,98 @@ class Full(NEVP):
...
-                    raise ValueError(
-                        "a_sparsity must be a tuple of 2D arrays.",
-                    )
-                if a.shape[0] != a.shape[1]:
-                    raise ValueError(
-                        "a_sparsity must be a tuple of square arrays.",
+                if self.zero_indices is None:
+                    warnings.warn(
+                        "No columns are zero in the first and last blocks. "
+                        "Reduction has no effect.",
```
almaeder: Shouldn't the warning be done only by one rank? There could be divergent sparsity between the ranks, but having every rank print a warning clutters the output.
Would the extra effort to synchronize between the ranks be worth it?

almaeder (changes requested):

src/qttools/nevp/full.py
vetschn: Changing this to be onlz on rank 0.

almaeder (approved):
Thank you for the work :shipit:

PR #250, opened 2026-03-11
Make the reference tests use a tmp_path to write outputs.

### LINKED ISSUE #219: Temporary folder for reference/example tests
Currently, the example reference tests write into the output folders in examples, but this has multiple problems:
- [ ] **Stale Data:** Old files are not deleted and can locally false flag tests
- [ ] **Race Conditions:** Running test with multiple workers can lead to tests overwriting each other's output
Better would be to use a temporary folder for each reference test.

Additionally, the following points should be addressed:
- [x] **Descriptive Naming:** Reference test naming: Currently, naming is `exampleX` which is not descriptive and makes debugging harder.
- [ ] **File Discovery:** The files in the new output folder are `glob`, but this can lead to reference files getting missed. The union of new and reference files should be tested to avoid this.
- [ ] **Error Printing:** Currently, the test stops at the first file it `glob`, but it would be better to reference test all files and afterwards give an error message of all files which were incorrect. This would allow better debugging.

vetschn:
To do this i added a function that resolves the input paths correctly. The parse_config function should not be called when modifying these paths, due to all the side effects. In the distributed examples it even leads to validation errors currently since the config is then not parsed in an MPI context.

Instead of this, use simple string replacement.

Closes #219

### DIFF
--- a/tests/quatrex/reference/test_reference.py
+++ b/tests/quatrex/reference/test_reference.py
@@ -1,12 +1,47 @@
 # Copyright (c) 2024-2026 ETH Zurich and the authors of the quatrex package.
 import subprocess
+import tomllib
 from pathlib import Path
 
 import numpy as np
 import pytest
 
 
-def test_single_rank(example: tuple[Path, bool]):
+def _adjust_config_paths(quatrex_config_path: Path, tmp_config_path: Path):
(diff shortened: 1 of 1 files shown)


almaeder (approved):
Great works 👍

Issue #222, opened 2026-02-03
Refactor data structures


almaeder:
There are some tings that are not ideal with the data structure. See sub-issues.

almaeder:
#333 fixed the last issues.

PR #252, opened 2026-03-13
Refactor the input loading in QTBM and SCBA

almaeder:
Currently, the input parsing for SCBA and QTBM is different:
- SCBA allows the construction from unit cell, but is complex.
- QTBM only allows the use of full Hamiltonians.
- Additionally, the symmetry is not used when loading the data.
This PR aims at simplifying and unifying the issues.
Furthermore, the symmetry will be used to lower the memory consumption during loading.


almaeder:
Currently, loading of `overlap` and `hamiltonian` is unified.
Points that should be added in later PRs:
- Not unsymmetrizing the matrix dict in QTBM
- Grid creation in QTBM for periodic devices
- Contact definition in both QTBM and SCBA

src/quatrex/device/device.py
```diff
@@ -123,43 +122,33 @@ def _init_hamiltonian(self) -> None:
...
             # Gamma point can be real depending on the basis.
+            # TODO: Do not unsymmetrize
             if not all(index == 0 for index in r):
-                self.hamiltonians[r] = self.hamiltonians[r].astype(xp.complex128)
-
-            if not self.hamiltonians[r].has_canonical_format:
-                self.hamiltonians[r].sum_duplicates()
-                self.hamiltonians[r].sort_indices()
+                self.hamiltonians[r] = hamiltonians[r].astype(xp.complex128)
+                # TODO: Having or not having the `sparse.csr_matrix` makes a difference
```
vetschn: In what way?

src/quatrex/core/config.py
vetschn, line 74: Do you want to change the default here while you're at it?

vetschn (commented):
Looks very reasonable to me!

src/quatrex/device/device.py
almaeder: I haven't investigated further, but I believe the issue was that `.T` gives a view and later code relies on accessing the underlying data vector of the sparse matrix.

src/quatrex/core/config.py
almaeder: There is an issue that, for MoS2, the reference checks do not hold on this branch or in dev when symmetry is True.

@awinka, do you have a guess why?

almaeder: Related, we should include reference tests for the same device with and without symmetry to avoid this.

almaeder:
Another point for a later PR:
Include an example for QTBM with "from unit cell".
The grid creation needs to be updated first,
while the contacts could still be given in real space.

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -997,10 +997,24 @@ def __iadd__(self, other: "DSDBSparse | sparse.spmatrix") -> "DSDBSparse":
...
+                            getattr(self._dsdbsparse, "cols"),
+                        ]
+                    )
+                )
+                return self._dsdbsparse
+
+            else:
+                raise NotImplementedError(
+                    "In-place addition only supported for dsdbcoo matrices"
+                )
```
vetschn, line 1017: This is a bit of a strange way to effectively deprecate the in-place addition for DSDBCSR. I think you could explicitly move the method out of the base class and only implement it for `dsdbcoo`.

src/quatrex/core/config.py
awinka: How large is the discrepancy?

src/qttools/datastructures/dsdbsparse.py
almaeder: Fair. The problem is that this is a method of `_DStackView`.

src/quatrex/device/device.py
```diff
@@ -123,43 +122,33 @@ def _init_hamiltonian(self) -> None:
...
+                self.hamiltonians[tuple(-i for i in r)] = sparse.csr_matrix(
+                    hamiltonians[r].copy().astype(xp.complex128).conj().T
+                )
+            else:
+                self.hamiltonians[r] = hamiltonians[r] + hamiltonians[r].conj().T
+                self.hamiltonians[r].setdiag(self.hamiltonians[r].diagonal() / 2)
```
awinka: Why do you only divide the diagonal by two, and not the full matrix?

almaeder: Because after loading the Hamiltonian, only the upper part of the 0,0,0 cell is kept.
Keeping half the x,y,z tuple and half the 0,0,0 is the minimum if there are no other symmetries.

almaeder: For QTBM, this change still needs to be propagated, but in the case of SCBA, the `triu` inside the `dsdbcoo` constructor will do nothing.

src/qttools/datastructures/dsdbsparse.py
almaeder: Referencing https://github.com/quatrex/quatrex/issues/222 here as well.

src/quatrex/device/inputs.py
```diff
@@ -457,96 +418,181 @@ def _create_matrix_from_unit_cells(
...
+            )
 
-    for coord, matrix in matrices.items():
-        unit_cells[coord] = xp.asarray(matrix).astype(xp.complex128)
+    # drop half the keys
+    matrix_dict = {
+        coord: matrix
+        for coord, matrix in matrix_dict.items()
+        if coord > (0, 0, 0) or (coord == (0, 0, 0))
+    }
```
awinka, line 561: Doesn't this drop more than half the keys?

src/quatrex/device/inputs.py
```diff
@@ -457,96 +418,181 @@ def _create_matrix_from_unit_cells(
...
+    if isinstance(matrix_dict[(0, 0, 0)], np.ndarray):
+        matrix_dict[(0, 0, 0)] = np.triu(matrix_dict[(0, 0, 0)])
+    elif isinstance(matrix_dict[(0, 0, 0)], sps.spmatrix):
+        matrix_dict[(0, 0, 0)] = sps.triu(matrix_dict[(0, 0, 0)])
+    else:
+        raise ValueError(
+            f"Matrix at coordinate (0,0,0) has unsupported type {type(matrix_dict[(0, 0, 0)])}."
+        )
+
+    # drop keys out side the neighbor cell cutoff if requested
```
awinka: out side -> outside

src/quatrex/device/inputs.py
almaeder: Maybe I should have written it differently.
Due to tuple comparisons being lexicographical, this should keep half. 
i.e. `(1,-1) > (0,0)` is true while `(-1,1) > (0,0)` is false.

src/quatrex/core/config.py
almaeder: Discrepancy is high. The observables are correct in the first iteration, but wrong in the second.
It could be in `_symmetrize_sigma`.

almaeder: Opened an issue #278 and will investigate.

tests/qttools/datastructures/test_dsdbsparse.py
```diff
@@ -605,6 +605,9 @@ def test_iadd_coo_stack(
         if symmetry_type[0]:
             pytest.skip("Symmetric DSDBSparse does not support in-place addition.")
 
+        if dsdbsparse_type.__name__ == "DSDBCSR":
```
vetschn: If for whatever reason we end up implementing this, this test will sneakily fail. It should be asserted that this raises some exception

src/quatrex/device/inputs.py
```diff
@@ -457,96 +418,181 @@ def _create_matrix_from_unit_cells(
...
-    # directions. The number of periodic cells is determined by the
-    # shape of the unit cell data.
-    for periodic_shift in xp.ndindex(transverse_repetitions):
-        # Center the periodic shift around zero.
-        periodic_shift = tuple(
-            [ps - (us // 2) for ps, us in zip(periodic_shift, transverse_repetitions)]
-        )
+        # Do not expand multiple time in
+        # transport direction
+        if coord[transport_ind] > 0:
```
vetschn: I think the case that `coord[transport_ind] < 0` never actually occurs because this is cut away during loading. I think `coord[transport_ind] != 0` may still be safer.

vetschn (commented):
See the comments. I think it is still worth investigating whether keeping the `triu` part for all matrices instead of dropping half the hopping terms is better

tests/quatrex/device/test_inputs.py
```diff
@@ -5,245 +5,201 @@
...
+    grid = create_coordinate_grid(coords, transport_cell_size, transport_ind, lat_vecs)
+    assert grid.shape == (transport_cell_size * 10, 3)
+
+    transport_cell = [1, 1, 1]
+    transport_cell[transport_ind] = transport_cell_size
+    transport_cell = tuple(transport_cell)
+
+    for ind in xp.ndindex(transport_cell):
+        row_ind = xp.ravel_multi_index(xp.asarray(ind), transport_cell)
         assert xp.allclose(grid[row_ind * 10 : (row_ind + 1) * 10], 1 + xp.array(ind))
```
vetschn, line 92: Some weird stuff here, why parameterize if you hardcode some variables afterwards?

src/quatrex/device/inputs.py
```diff
@@ -1,250 +1,167 @@
...
-    )
-
-    return trimmed_matrix
-
-
 def create_coordinate_grid(
-    wannier_centers: NDArray,
-    supercell_size: tuple,
+    unit_cell_coords: NDArray,
+    transport_cell_size: int,
```
vetschn: This should be renamed to `num_unit_cells`, no?

src/quatrex/device/device.py
maurodossena: Comment is wrong

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

src/quatrex/device/device.py
```diff
@@ -123,43 +122,33 @@ def _init_hamiltonian(self) -> None:
...
-
-            self.hamiltonians[r] = sparse.csr_matrix((h_r))
+        # NOTE: Contains only the upper triangular parts of the matrices.
+        hamiltonians = load_matrices(self.config, "hamiltonian")
 
+        self.hamiltonians = {}
+        for r, h_r in hamiltonians.items():
             # TODO: Check data type handling.
             # Gamma point can be real depending on the basis.
+            # TODO: Do not unsymmetrize
```
awinka, line 132: Why do you unsymmetrize?

src/quatrex/device/device.py
```diff
@@ -123,43 +122,33 @@ def _init_hamiltonian(self) -> None:
...
             if not all(index == 0 for index in r):
```
awinka: Do you need this condition? Shouldn't it also work for r=(0,0,0)?

src/quatrex/device/inputs.py
```diff
@@ -86,235 +83,153 @@ def load_structure(
...
+            coord = list(shift)
+            coord[transport_ind] += r_j - r_i
+            coord = tuple(int(i) for i in coord)
+
+            flipped_coord = tuple(-int(i) for i in coord)
+
+            if coord in matrix_dict:
+                block = matrix_dict[coord]
+                block_flipped = matrix_dict[flipped_coord]
+                block = block + xp.triu(block_flipped, k=1).conj().T
```
awinka, line 182: This doesn't have any special handling if `block=block_flipped` as in `_init_hamiltonian` line 133. If it is not needed, could the special handling in `_init_hamiltonian` be removed (see comment in line 133 in file `device.py`)?

awinka (commented):
Looks solid! I just have two comments below

src/quatrex/device/device.py
almaeder: Because QTBM is not adapted yet for symmetry.

src/quatrex/device/device.py
almaeder: True, it is not needed and will be removed 👍.
I didn't catch this.

src/quatrex/device/inputs.py
almaeder: The special handling in `_init_hamiltonian` for QTBM was a leftover from when half the keys were dropped instead of keeping only the upper part of all matrices.

src/quatrex/device/inputs.py
```diff
@@ -86,235 +83,153 @@ def load_structure(
...
-    supercell_size: tuple,
+    unit_cell_coords: NDArray,
+    num_unit_cells: int,
+    transport_ind: int,
     lattice_vectors: NDArray,
 ) -> NDArray:
-    """Creates a grid of coordinates for Wannier centers in a supercell.
+    """Creates a grid of coordinates for orbital centers in a transport cell.
+    Only expands in the transport direction, and assumes that the unit cell is
+    repeated in the transport direction.
```
vetschn: I think here the description is a bit misleading.
```suggestion
def create_coordinate_grid(
    unit_cell_coords: NDArray,
    num_unit_cells: int,
    transport_ind: int,
    lattice_vectors: NDArray,
) -> NDArray:
    """Creates a grid of coordinates for orbital centers in a supercell.
    
    This takes an array of coordinates and repeats them `num_unit_cells`
    times along the transport direction, by shifting them by the lattice
    vector in that direction.
```

src/quatrex/device/inputs.py
```diff
@@ -419,17 +328,49 @@ def _tile_sparse_blocks(block, num_blocks, offsets):
...
+    matrix_dict : dict[tuple, sparse.csr_matrix | NDArray]
+        The dictionary of matrices corresponding to different periodic
+        repetitions. It is assumed that only the upper parts are present.
+    symmetric : bool
+        Whether the resulting matrix should be symmetric. If `True`, only
+        construct the upper triangular part.
+    phases :
+
+    """
+
```
vetschn:
```suggestion
    """Sums up periodic image contributions for a specific k-point.
    
    This takes a Hermitian operator (e.g., Hamiltonian, overlap) and performs a weighted
    sum to construct a matrix for a specific k-point.

    Parameters
    ----------
    matrix_dict : dict[tuple, sparse.csr_matrix | NDArray]
        The dictionary of matrices corresponding to different periodic
        repetitions. It is assumed that only the upper parts are present.
    symmetric : bool
        Whether the resulting matrix should be symmetric. If `True`, only
        construct the upper triangular part.
    phases : dict, optional
        A dictionary mapping the different image indices to their weight.
        If not provided, this performs an unweighted sum.

    """

```

src/quatrex/device/inputs.py
```diff
@@ -527,96 +465,169 @@ def _create_matrix_from_unit_cells(
...
+    elif isinstance(matrix_dict[(0, 0, 0)], sps.spmatrix):
+        matrix_dict = {
+            coord: sparse.csr_matrix(matrix).astype(xp.complex128)
+            for coord, matrix in matrix_dict.items()
+        }
+
+    # expand potentially if the system is periodic
+    # and given bz unit cell matrix_dict
+    if config.device.construct_from_unit_cell:
+        # raise NotImplementedError("Constructing from unit cell is not implemented yet.")
```
vetschn: Should this still be here?
```suggestion
```

vetschn (changes requested):
See the small suggested changes. Could you maybe rename the `quatrex/tests/quatrex/banstructure` folder to `quatrex/tests/quatrex/ban*d*structure` as well in this PR?

vetschn (approved):

PR #253, opened 2026-03-13
move hilbert transfrom from global function to private class function…

YongDaLi:
… in class `PCoulombScreening`. The transform is not used anywhere else, so it doesn't need to be accessible/shareable. 

Also ran `black` formatting which removed some empty lines.

### DIFF
--- a/src/quatrex/coulomb_screening/polarization.py
+++ b/src/quatrex/coulomb_screening/polarization.py
@@ -18,45 +18,6 @@ if xp.__name__ == "cupy":
     cache = xp.fft.config.get_plan_cache()
 
 
-def hilbert_transform(a: NDArray, energies: NDArray) -> NDArray:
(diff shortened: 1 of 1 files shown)


YongDaLi:
Okay, turns out `hilbert_transform` is used 2 places:
1. `polarization.py`, called in class `PCoulombScreening`
2. `sse_coulomb_screening.py`, called in class `SigmaCoulombScreening`

But they're defined differently. The first one is `a` Hilbert transform of a. But the second one is a Hilbert transform `sg - sl`. 

Maybe there's a way to uniform them and improve code re-use?

YongDaLi:
I'll deal with this later when (or if) I merge my `adaptive_energy_grids_dev` branch with `dev`. When performing the FFT on the uniform "filled-in" interpolated grid, we need to pass in the new "filled-in" spacing to create the Hilbert kernel. Since there's code re-use, I moved this out to another function.

Issue #256, opened 2026-03-18
If coulomb_screening=false, the code crashes when checking non-existent GW objects


Kosheira1:
```
│ /home/scratch.ldeuschle_other/miniforge/envs/quatrex-dev/lib/python3.13/site-packages/qttools/profiling/profiler.py:342 in wrapper                                                   │
│                                                                                                                                                                                      │
│   339 │   │   │   │                                                                                                                                                                  │
│   340 │   │   │   │   # Call the function.                                                                                                                                           │
│   341 │   │   │   │   with self.profile_range(label, level, comm):                                                                                                                   │
│ _ 342 │   │   │   │   │   result = func(*args, **kwargs)                                                                                                                             │
│   343 │   │   │   │                                                                                                                                                                  │
│   344 │   │   │   │   return result                                                                                                                                                  │
│   345                                                                                                                                                                                │
│                                                                                                                                                                                      │
│ /home/scratch.ldeuschle_other/miniforge/envs/quatrex-dev/lib/python3.13/site-packages/quatrex/core/scba.py:434 in _symmetrize_sigma                                                  │
│                                                                                                                                                                                      │
│   431 │   │   │   # Make the self-energy Hermitian (removing the skew-Hermitian part).                                                                                               │
│   432 │   │   │   self.data.sigma_retarded.symmetrize(xp.add)                                                                                                                        │
│   433 │   │                                                                                                                                                                          │
│ _ 434 │   │   if self.config.coulomb_screening.discard_real_parts:                                                                                                                   │
│   435 │   │   │   self.data.sigma_lesser._data.real = 0                                                                                                                              │
│   436 │   │   │   self.data.sigma_greater._data.real = 0                                                                                                                             │
│   437 │   │   │   # Make sure that the imaginary part comes only from                                                                                                                │
_──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────_
AttributeError: 'NoneType' object has no attribute 'discard_real_parts'
```

vetschn:
Fixed in #257. Whether or not spurious, non-axis-aligned components of the self-energy should be discarded is now controlled with the `scba.align_self_energy_to_complex_axes` keyword.

Issue #254, opened 2026-03-18
`compute_retarded_polarization` is a misleading flag name


YongDaLi:
In `config.py` under CoulombScreeningConfig, the flag name `compute_retarded_polarization` is confusing.
https://github.com/quatrex/quatrex/blob/bb10b3d52085da3d624606f80953541d8ce7062b/src/quatrex/core/config.py#L549-L555

Even if you set the flag to `False`, Quatrex will still compute the retarded part of the Polarization. But it just computes it as $(P^{>} - P^{<})/2$.

https://github.com/quatrex/quatrex/blob/bb10b3d52085da3d624606f80953541d8ce7062b/src/quatrex/coulomb_screening/polarization.py#L228-L237

I think there's 2 methods/formalisms to compute P retarded. Ideally the flag name should specify which method to compute the Polarization and then how much of the method is actually implemented.

Or if you just want to rename the variable, it should be something like `use_retarded_polarization_approximation`.

almaeder:
@YongDaLi this is fixed by #257 right?

YongDaLi:
Yes, it's resolved

PR #257, opened 2026-03-19
Fix discard parameters

almaeder:
This PR addresses two open issues: #254 and #256.
This is done by splitting the parameters and renaming them.

### DIFF
--- a/src/quatrex/core/config.py
+++ b/src/quatrex/core/config.py
@@ -72,6 +72,16 @@ class SCBAConfig(BaseModel):
 
     symmetric: bool = False
 
+    use_sigma_approximation: bool = True
+    r"""Whether to discard parts of the self-energies.
+
+    This affects the self-energies in the following way:
+    The real parts of the lesser/greater self-energies are discarded
+    and the imaginary part of retarded self-energy computed only from
+    the lesser and greater parts by $\frac{\mathbf{\Sigma}^> - \mathbf{\Sigma}^<}{2}$.
+
+    """
+
 
 class PoissonConfig(BaseModel):
     """Options for the Poisson solver."""
--- a/src/quatrex/core/config.py
+++ b/src/quatrex/core/config.py
@@ -72,13 +72,18 @@ class SCBAConfig(BaseModel):
 
     symmetric: bool = False
 
-    use_sigma_approximation: bool = True
-    r"""Whether to discard parts of the self-energies.
+    align_self_energy_to_complex_axes: bool = True
+    r"""Whether to discard parts of the self-energy.
 
-    This affects the self-energies in the following way:
-    The real parts of the lesser/greater self-energies are discarded
-    and the imaginary part of retarded self-energy computed only from
-    the lesser and greater parts by $\frac{\mathbf{\Sigma}^> - \mathbf{\Sigma}^<}{2}$.
+    This affects the self-energy in the following way:
+    - The real parts of the lesser/greater self-energy are discarded.
+    - The imaginary part of the retarded self-energy from any previous
+    computation is zeroed.
+
+    This happens before the imaginary part of the retarded self-energy
+    is computed from the lesser and greater parts as
+    $$\mathrm{Im}\left[\mathbf{\Sigma}^R\right] =
+    \frac{\mathbf{\Sigma}^> - \mathbf{\Sigma}^<}{2i}$$.
 
     """
 
@@ -112,7 +117,7 @@ class MemoizerConfig(BaseModel):
 
     mode: Literal["auto", "force", "force-after-first", "off"] = "auto"
     """The memoization mode to determine when to do fixed-point iterations.
-    
+
     - "auto": Automatically decides whether to use memoization based on the
         specified tolerances. Only useful if all ranks memoize.
     - "force": Always use memoization.
@@ -125,7 +130,7 @@ class MemoizerConfig(BaseModel):
 
     relative_tol: PositiveFloat = 2e-1
     """The relative tolerance for the fixed-point iterations.
-    
+
     Only used if `mode` is set to "auto".
     """
 
@@ -172,7 +177,7 @@ class OBCConfig(BaseModel):
 
     algorithm: Literal["sancho-rubio", "spectral"] = "spectral"
     """The OBC algorithm to use.
-    
+
     - "sancho-rubio": Uses the Sancho-Rubio iterative scheme to compute the
         surface Green's functions. This method achieves exponential convergence
         compared to the linear convergence of fixed-point iterations.
@@ -231,14 +236,14 @@ class OBCConfig(BaseModel):
     $$ \mathbf{g}_{n+1} = [\mathbf{M}_{0} - \mathbf{M}_{-1} \mathbf{g}_{n} \mathbf{M}_{1} ]^{-1} $$
 
     This is needed to improve the accuracy of the surface Green's functions
-    if not enough eigenpairs are considered. 
+    if not enough eigenpairs are considered.
 
     Only used if `algorithm` is set to "spectral".
     """
 
     min_propagation: PositiveFloat = 1e-2
     r"""The minimum propagation speed for propagating modes.
-    
+
     The propagation speed is computed as:
 
     $$ abs(real(\frac{dE}{dk})) / abs(imag(\frac{dE}{dk})) $$
@@ -247,7 +252,7 @@ class OBCConfig(BaseModel):
 
     residual_tolerance: PositiveFloat = 1e-3
     r"""The tolerance for the residual of the eigenpairs.
-    
+
     The residuals are computed as:
 
     $$ \lvert \sum \limits_{n=-b}^{b} \lambda^{b} \mathbf{M}_{n} \vec{v} \rvert $$
@@ -259,7 +264,7 @@ class OBCConfig(BaseModel):
 
     residual_normalization: bool = True
     """Whether to normalize the residuals by the norm of the eigenvalue.
-    
+
     This is useful to avoid that large eigenvalues have large residuals
     and small eigenvalues have small residuals.
     """
@@ -271,7 +276,7 @@ class OBCConfig(BaseModel):
     The residual is computed as:
 
     $$ \lvert \mathbf{g} - [\mathbf{M}_{0} - \mathbf{M}_{-1} \mathbf{g} \mathbf{M}_{1} ]^{-1} \rvert / \lvert \mathbf{g} \rvert $$
-    
+
     This parameter is only used if the `formalism` is `wf`. Otherwise, the memoizer
     is responsible for residual checking and warnings.
     """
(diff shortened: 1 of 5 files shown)


src/quatrex/coulomb_screening/polarization.py
```diff
@@ -84,8 +84,10 @@ def __init__(
         )
         self.batch_size = config.compute.convolve.batch_size
 
-        self.discard_real_parts = config.coulomb_screening.discard_real_parts
-        self.compute_retarded = config.coulomb_screening.compute_retarded_polarization
+        self.use_approximation = config.coulomb_screening.use_polarization_approximation
```
YongDaLi: Can we propagate the variable name to the `PCoulombScreening` class as well?

i.e.
```python
self.use_polarization_approximation = config.coulomb_screening.use_polarization_approximation
```

I can imagine myself getting super confused if I'm looking at line 234 and see `if self.use_approximation`. What approximation??

YongDaLi (changes requested):
The other `quatrex_config.toml` files need to be updated as well, at some point. Because they now have flags that don't exist, so the config validation will complain. But I think it should be up to individual affected users to update them. If we update all of them now, then a rebase from their branch onto dev will break their existing config files.

So other than the variable name propagation, looks good to me

src/quatrex/coulomb_screening/polarization.py
almaeder: The idea is that the name `polarization` does not need to be repeated since it is implied by the class/file name.
The name in the config is already a bit wordy, but there is not (yet) a polarization config inside `CoulombScreeningConfig`.

YongDaLi: Okay fine. I was convinced that the variable name is acceptable.

src/quatrex/core/config.py
```diff
@@ -536,17 +546,17 @@ class CoulombScreeningConfig(BaseModel):
...
-    the imaginary part is only computed from only the lesser and greater parts by $\frac{\mathbf{X}^> - \mathbf{X}^<}{2}$.
-    Else, the imaginary part can also contain contributions from the Hilbert transformation.
+    use_polarization_approximation: bool = True
+    r"""Whether to discard parts of the polarization.
 
+    This affects the polarization in the following way:
+    The real parts of the lesser/greater polarization are discarded. 
+    The imaginary part of the retarded polarization from previous computation is zeroed.
+    The retarded polarization is computed only using lesser and greater parts by $\frac{\mathbf{P}^> - \mathbf{P}^<}{2}$.
+    Thus, the retarded polarization is ensured to have zero real part and only an imaginary part.
```
YongDaLi: Following my chat with @almaeder, we agree the most important thing is to have a clear description in the config so Quatrex users can easily understand. 

I've updated the config description to be super explicit in describing what the code does, with the main point of the approximation flag being `Thus, the retarded polarization is ensured to have zero real part and only an imaginary part.`

vetschn:
I am maybe a bit confused what the logic should be here. I think there should be only one keyword like `include_energy_renormalization`, that controls whether or not the principal value of that integral is included in the calculation, because that is what that integral actually does, right?

I don't understand why you need both `compute_hilbert_retarded` and `use_approximation`. The `use_approximation` parameter seems to just control whether the quantities are made purely real / imaginary, no?

almaeder:
> I am maybe a bit confused what the logic should be here. I think there should be only one keyword like `include_energy_renormalization`, that controls whether or not the principal value of that integral is included in the calculation, because that is what that integral actually does, right?
> 
> I don't understand why you need both `compute_hilbert_retarded` and `use_approximation`. The `use_approximation` parameter seems to just control whether the quantities are made purely real / imaginary, no?

Yes, the parameters control different things since even if you include the hilbert correction, the approximation of keeping only real/imaginary part could be done. How the things should be named :shrimp:.
Then there is the third parameter `apply_hilbert_correction` which controls if the hilbert part is used if there should be an additional correction due to the finite energy grid.
We should include all three parameters, but the names should change.

We could do:
`discard_imaginary_parts`
`include_hilbert_part`
`correct_hilbert_part` (only matters if hilbert is included)
and put them into a separate polarization config.

vetschn:
Okay once again, these parameters do not really make sense. To summarize what is happening now (mainly for myself):
- We have one parameter (`use_*_approximation`) to control whether to discard the real part of the lesser/greater polarization/self-energy
- We have another parameter (`compute_hilbert_retarded_polarization`) that controls whether to include the purely real, energy-renormalizing principal value of that integral but only in the polarization. There is no option to turn this off for the retarded self-energy.
- We have another parameter (`apply_hilbert_correction`) for the coulomb screening self-energy, that controls whether to correct the energy-grid edges.


Now what i think we should have:
- `include_*_energy_renormalization` for both `p_retarded` and `sigma_retarded`, just controlling whether or not to include that principal value. I guess for sigma we always want the printegral on, right @awinka?
- `align_*_to_complex_axes` that controls whether or not to force lesser/greater components to be aligned with the imaginary axis, and to make sure that the retarded part is purely real up to that point. Then it's crucial that this alignment happens before `*_retarded += (*_greater - *_lesser) / 2`.
- And we keep something like `correct_energy_renormalization_edges` for the coulomb screening self-energy.

What is maybe also a bit strange is that the `*_retarded += (*_greater - *_lesser) / 2` part for the self-energy is hidden away in the `symmetrize_sigma` method of the SCBA class.

vetschn:
@almaeder could you maybe check whether this looks acceptable, or more confusing now?

src/quatrex/core/scba.py
```diff
@@ -793,6 +788,11 @@ def run(self) -> None:
             # Symmetrize the self-energy.
             self._symmetrize_sigma()
 
+            # Add the anti-Hermitian part to the retarded self-energy.
+            self.data.sigma_retarded.data += 0.5 * (
+                self.data.sigma_greater.data - self.data.sigma_lesser.data
+            )
+
```
almaeder, line 795: It makes more sense to do this here than in `_stash_sigma`

almaeder:
> @almaeder could you maybe check whether this looks acceptable, or more confusing now?

The changes make sense to me. A good change to also make the energy renormalization optional for the SEE self-energy.
In a later PR, we should refactor the branching in the SEE self-energy since lesser/greater have now two "different" ways in which they are calculated (but they should do the same, supposedly).

awinka:
> Now what i think we should have:
> 
>  * `include_*_energy_renormalization` for both `p_retarded` and `sigma_retarded`, just controlling whether or not to include that principal value. I guess for sigma we always want the printegral on, right @awinka?

I think it is good that the principle integral can be an option for sigma too, in case you only want to study the effects of scattering. The default for GW calculations should be that it is turned on for sigma, since the bandgap opening is one of the main results of GW.

src/quatrex/core/config.py
```diff
@@ -72,6 +72,21 @@ class SCBAConfig(BaseModel):
 
     symmetric: bool = False
 
+    align_self_energy_to_complex_axes: bool = True
```
YongDaLi, line 75: Should be align to `imag_axis`, because `complex_axes` is technically both real and imag axes.

src/quatrex/core/config.py
```diff
@@ -72,6 +72,21 @@ class SCBAConfig(BaseModel):
...
+    This affects the self-energy in the following way:
+    - The real parts of the lesser/greater self-energy are discarded.
+    - The imaginary part of the retarded self-energy from any previous
+    computation is zeroed.
+
+    This happens before the imaginary part of the retarded self-energy
+    is computed from the lesser and greater parts as
+    $$\mathrm{Im}\left[\mathbf{\Sigma}^R\right] =
+    \frac{\mathbf{\Sigma}^> - \mathbf{\Sigma}^<}{2i}$$.
+
```
YongDaLi, line 87: I would add a clarifying line at the end this docstring. Something like: "This ensures that the retarded polarization only contains an imaginary part and no real part".

src/quatrex/core/config.py
```diff
@@ -529,31 +544,73 @@ class CoulombScreeningConfig(BaseModel):
...
+    \mathbf{P}^<}{2i}$$.
 
-    Computing the correction is slightly more expensive.
     """
 
-    discard_real_parts: bool = True
-    r"""Whether to discard the real parts of the lesser/greater polarization and self-energy.
+    include_energy_renormalization: Literal["self-energy", "polarization", "both"] = (
+        "self-energy"
+    )
```
YongDaLi, line 564: Exposing the settings like this makes sense to me, because it's explained from a physics perspective rather than a computational perspective. I support.

src/quatrex/core/config.py
vetschn: That would not be correct. The retarded self-energy can still contain real part. Only any spurious complex part is discarded.

src/quatrex/core/config.py
vetschn: That's the point of the name. Lesser and greater are aligned to the imaginary axis and retarded is aligned to real axis.

src/quatrex/core/config.py
YongDaLi: Ah my bad, sorry. This should be uner the polarization part.

src/quatrex/core/config.py
```diff
@@ -529,31 +544,73 @@ class CoulombScreeningConfig(BaseModel):
...
+
+    This affects the polarization in the following way:
+    - The real parts of the lesser/greater polarization are discarded.
+    - The imaginary part of the retarded polarization from previous
+    computation is zeroed.
+
+    This happens before the imaginary part of the retarded polarization
+    is computed from the lesser and greater parts as
+    $$\mathrm{Im}\left[\mathbf{P}^R\right] = \frac{\mathbf{P}^> -
+    \mathbf{P}^<}{2i}$$.
```
YongDaLi, line 558: copy-paste from self-energy part (I meant to write it here): I would add a clarifying line at the end this docstring. Something like: "This ensures that the retarded polarization only contains an imaginary part and no real part".

vetschn: Also not the case here, sorry 😅 Real part of the retarded polarization is controlled by the `include_energy_renormalization` parameter. This is only to suppress non-axis-aligned spurious components

src/quatrex/core/config.py
YongDaLi: I see. Still feels like a convoluted name. I think the root problem is that we're trying to say a lot in 1 variable name. I can't immediately think of a better word than "align". It's better than the previous "approximate". 

I suppose it's good enough. The docstring is quite descriptive.

src/quatrex/core/config.py
YongDaLi: nevermind, I see that the real part of the polarization depends on the Hilber transform.

YongDaLi (approved):
I feel the config name `align_x_to_complex_axes` is still confusing on first look. But with the docstring, it's sufficient.

PR #265, opened 2026-03-26
Update Coulomb matrix, reference outputs and configuration for CNT GW examples

### LINKED ISSUE #264: Unclear where the current CNT Coulomb matrix comes from
The Coulomb matrix for the GW examples is not computed in the point charge approximation. I was unable to figure out where the current matrix came from exactly. The magnitude of the values is drastically different.

This should be changed, just to be consistent with the unit-cell inputs. The epsilon factor and the reference output data will have to be adjusted as well.

While checking this i also noticed that the carbon-nanotube unit-cell inputs come from a different Wannierization than the ones in the "regular" full matrix inputs. Both wannierizations reproduce the bandstructure perfectly and identically (`np.linalg.norm(e_k - e_k_unit_cells) / np.linalg.norm(e_k_unit_cells)` is 0.1% for the whole Brillouin zone, at gamma it is 0.01%, so the exact same band edges). This is therefore not really a problem.

vetschn:
The previous Coulomb matrix was not in the point-charge approximation. Fixed this. To get converging simulations, i had to set the epsilon_r parameter to 4 as is the case in the unit-cell example.

Since the reference tests are run both in distributed and in single-rank mode, small random variations can occur between the two cases when using Beyn. Switched to the full NEVP solver to try to fight this.

This will close #264


vetschn:
@almaeder I am also trying to fix the conditions under which the reference tests are run in the current setup. The single rank tests are run in both distributed and single rank mode, which is very unexpected to me.

The distributed tests are run in a way where each rank spawns another six ranks. I think that makes no sense and is wasting CI time, especially on a 4-core runner.

Or do you disagree? Is there a reason for this?

EDIT:

I had misunderstood what was actually happening, but i'd say what's happening is still not really expected. All tests were always ran in an MPI context. Removed the `--with-mpi` flag from the pytest invocation to stop this.

examples/w90/carbon-nanotube/gw-dist/quatrex_config.toml
```diff
@@ -57,7 +57,7 @@ solver.max_batch_size = 1000
 # solver.compute_current = true # Compute Meir-Wingreen current.
 
 obc.algorithm = "spectral"
-obc.nevp_solver = "beyn"
+obc.nevp_solver = "full"
 obc.block_sections = 1
```
almaeder, line 61: Makes sense to switch to full to avoid the randomness.

vetschn:
Reference tests still don't pass with the new coulomb matrix. They are close but not within the currently set tolerance. I'll first try further increasing the `epsilon_r` value and then i might have to increase the tolerances.

.github/workflows/tests.yaml
```diff
@@ -41,15 +41,15 @@ jobs:
       - name: Run NUMBA Tests (QTTools Kernels)
         if: success() || failure()
         run: |
-          coverage run --source=src --parallel-mode -m pytest --with-mpi tests/qttools/kernels
```
almaeder, line 44: Certain tests do not run anymore due to the removal of this.
Only a few, which are associated with the data structure.
We concluded to duplicate these tests to have a clear separation.

examples/w90/carbon-nanotube/gw/quatrex_config.toml
```diff
@@ -82,7 +83,7 @@ solver.max_batch_size = 1000
 
 obc.algorithm = "spectral"
 obc.nevp_solver = "beyn"
-obc.block_sections = 2
+obc.block_sections = 1
```
almaeder, line 86: Why was this changed if `num_connected_blocks = 2`?

vetschn: I think the change to the point-charge-approximated Coulomb matrix causes a bit stronger screening. My guess is that the variation of the self-energy in the two blocks close to the leads is still quite large and ends up causing spurious OBC when using block-sectioning. Reducing this fixed things and also finally made the tests pass.

Should i look into this again?

almaeder: No, it is fine. Also, it is the example with the "less" constant potential in the contact, right?

examples/w90/carbon-nanotube/gw-dist/quatrex_config.toml
vetschn: I think it should still be okay since the RNG is actually seeded properly. I think it should even be possible to go back to Beyn here as well and still have the tests pass.

examples/w90/carbon-nanotube/gw/quatrex_config.toml
vetschn: > Also, it is the example with the "less" constant potential in the contact, right?

Not so sure about this, i think it was more the interaction strength with the new input

.github/workflows/tests.yaml
almaeder: Edit: not duplication, but having a child class, which has mark mpi, while the parent does not.

tests/quatrex/reference/test_reference.py
```diff
@@ -41,6 +41,9 @@ def _adjust_config_paths(quatrex_config_path: Path, tmp_config_path: Path):
     tmp_config_path.write_text(config_text)
 
 
+# NOTE: Skip this if running in an MPI environment. These should be run
+# in a single process only.
+@pytest.mark.mpi_skip()
```
almaeder, line 46: Note: This is not 100% needed for the CI, as it uses `--only-mpi`, but we recommend using this since a user could still run with `--with-mpi`. 
`--with-mpi` always opens an MPI environment even if there is a single rank.

almaeder (approved):
LGTM

Issue #231, opened 2026-02-20
Restructure Pydantic configuration models


vetschn:
The config options could use some refactoring.
- Perhaps split certain models up into different modules? This would mean another subpackage `quatrex.config`.
- The compute options sometimes relate to very specific parts of the code (e.g. `NEVPConfig.eig_compute_location`). We could reshuffle these directly into the other  simulation options (so move this to `OBCConfig.nevp.eig_compute_location` for instance) for more fine-grained control.

almaeder:
- Maybe avoid transformation to numpy arrays in validators.

almaeder:
Further hierarchies are needed in the `OBC` config since now all parameters for different OBC and eigenvalue solvers are together.

PR #266, opened 2026-03-30
Fix Meir-Wingreen current computations in RGF

### LINKED ISSUE #262: Perform NEGF boundary current computations with full blocks
The computation of boundary currents is currently done after the blocks of the Green's functions have already been reduced to whatever is in the sparsity mask.

As we anyways already have the option for returning the Meir-Wingreen device current from the RGF solver, we should extend this to the boundary currents.

This unfortunately goes against the idea of separating the solvers from the physics (sorry @vincent-maillou, also related to #67 ) but it is obviously more important to get accurate currents from the simulator.

I think boundary currents could even be worked into the `serinv` solver backend, but that will have to be checked.

### LINKED ISSUE #271: Meir-Wingreen device current for distributed RGF
We need to check whether we can come up with a more general formulation for Meir-Wingreen that we can also use in the distributed solver to compute the device currents using the dense blocks. Either we can then include this in `RGFDist` or we can rule it out completely.

vetschn:
The current buffer returned from RGF now includes the current from and to the contacts at either end, as well as the current at each layer. It is now of size `num_blocks + 1` (before `num_blocks - 1`).

The computation of the Meir-Wingreen current had a bug as well. Fixed that.

The `compute_current` parameter was previously set to `False` by default. It is now set to `True` if `"rgf"` is being used. Only used for electrons.

Contact current computation in the observables is removed. It is better to do it in RGF. The corresponding output files are also gone. Now everything you need is the Meir-Wingreen current.

Contact current difference computation in SCBA also changed to use that instead.

This will close #262

### DIFF
--- a/src/qttools/greens_function_solver/rgf.py
+++ b/src/qttools/greens_function_solver/rgf.py
@@ -172,8 +172,10 @@ class RGF(GFSolver):
             obc_blocks = OBCBlocks(num_blocks=a.num_blocks)
 
         if return_current:
-            # Allocate a buffer for the current.
-            current = xp.zeros((*a.shape[:-2], a.num_blocks - 1), dtype=a.dtype)
+            # Allocate a buffer for the current. This includes current
+            # between each layer and from/to the leads (in total
+            # num_blocks + 1).
+            current = xp.zeros((*a.shape[:-2], a.num_blocks + 1), dtype=a.dtype)
 
         # Get list of batches to perform
         batches_sizes, batches_slices = get_batches(a.shape[0], self.max_batch_size)
@@ -379,6 +381,7 @@ class RGF(GFSolver):
                 )
 
                 if return_current:
+                    a_ji_dagger = a_ji.conj().swapaxes(-2, -1)
                     a_ji_xr_ii = a_ji @ xr_ii
                     a_ji_xr_ii_sx_ij = a_ji_xr_ii @ sigma_lesser_ij
                     sigma_lesser_tilde = (
@@ -392,7 +395,7 @@ class RGF(GFSolver):
                         + a_ji_xr_ii_sx_ij.conj().swapaxes(-2, -1)
                         - a_ji_xr_ii_sx_ij
                     )
-                    current[stack_slice, ..., i] = xp.trace(
+                    current[stack_slice, ..., j] = xp.trace(
                         sigma_greater_tilde @ xl_diag_blocks[j]
                         - xg_diag_blocks[j] @ sigma_lesser_tilde,
                         axis1=-2,
@@ -402,6 +405,22 @@ class RGF(GFSolver):
                 xr_diag_blocks[i] = xr_ii + xr_ii_a_ij_xr_jj_a_ji @ xr_ii
                 if return_retarded:
                     xr_.blocks[i, i] = xr_diag_blocks[i]
+
+        if return_current:
+            current[..., 0] = xp.trace(
+                obc_blocks.greater[0] @ xl_diag_blocks[0]
+                - xg_diag_blocks[0] @ obc_blocks.lesser[0],
+                axis1=-2,
+                axis2=-1,
+            )
+            # NOTE: Negative sign is needed to get the current flowing
+            # in the correct direction (positive from left to right).
+            current[..., -1] = -xp.trace(
+                obc_blocks.greater[-1] @ xl_diag_blocks[-1]
+                - xg_diag_blocks[-1] @ obc_blocks.lesser[-1],
+                axis1=-2,
+                axis2=-1,
+            )
 
         if out is None:
             if return_retarded:
--- a/src/quatrex/core/config.py
+++ b/src/quatrex/core/config.py
@@ -158,10 +158,42 @@ class SolverConfig(BaseModel):
     # The maximum number of energies per batch.
     max_batch_size: PositiveInt = 100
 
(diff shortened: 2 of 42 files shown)


vetschn:
Would be great to already have a review here @awinka / @almaeder. The reference example data and the configs still need to be updated. Waiting for that until #265 is merged.

src/qttools/greens_function_solver/rgf.py
```diff
@@ -379,6 +381,7 @@ def selected_solve(
                 )
 
                 if return_current:
+                    a_ji_dagger = a_ji.conj().swapaxes(-2, -1)
```
awinka, line 384: This was the bug?

awinka (approved):
Looks good to me!

src/qttools/greens_function_solver/rgf.py
vetschn: Indeed :bug:

src/quatrex/core/config.py
```diff
@@ -158,11 +158,43 @@ class SolverConfig(BaseModel):
     # The maximum number of energies per batch.
     max_batch_size: PositiveInt = 100
 
-    # Whether to compute the current via the Meir-Wingreen formula.
-    compute_current: bool = False
+    compute_current: bool | None = None
```
almaeder, line 161: Issue for the future: It is less intuitive that this parameter is in the `SolverConfig` and not in the `OutputConfig`.
The problem will be that the validator needs to know the `algorithm`.

src/quatrex/core/scba.py
```diff
@@ -463,11 +458,14 @@ def _has_converged(self) -> bool:
...
-        i_right = xp.real(self.observables.electron_current.get("right", 0.0))
+        meir_wingreen_current = self.observables.electron_current.get(
+            "meir-wingreen", [0, 0]
+        )
+        i_left = xp.real(meir_wingreen_current[0])
+        i_right = xp.real(meir_wingreen_current[-1])
 
         dE = self.electron_energies[1] - self.electron_energies[0]
-        current_diff = xp.abs(xp.sum(i_left) * dE + xp.sum(i_right) * dE)
+        current_diff = xp.abs(xp.sum(i_left) * dE - xp.sum(i_right) * dE)
```
almaeder, line 468: Another issue for the future: We should report both absolute and relative current differences. Also, we shouldn't only define it through the contact currents.

(This has to be revisited when there are more than two contacts)

src/quatrex/core/scba.py
```diff
@@ -463,11 +458,14 @@ def _has_converged(self) -> bool:
         max_diff = np.empty_like(local_max_diff)
         global_comm.Allreduce(local_max_diff, max_diff, op=MPI.MAX)
 
-        i_left = xp.real(self.observables.electron_current.get("left", 0.0))
-        i_right = xp.real(self.observables.electron_current.get("right", 0.0))
+        meir_wingreen_current = self.observables.electron_current.get(
+            "meir-wingreen", [0, 0]
+        )
+        i_left = xp.real(meir_wingreen_current[0])
+        i_right = xp.real(meir_wingreen_current[-1])
```
almaeder, line 466: Will this not lead to an issue with the distributed solver?
Maybe the dist solver can also have a fake Meir-Wingreen current with only the contact currents.

This could then be extended after we figure out the correct formulation in the distributed case.

src/quatrex/core/scba.py
vetschn: Opened a new issue #270

src/quatrex/core/scba.py
vetschn: Nope, if `compute_current` in the solver config is set to `true`, the distributed code will raise a `NotImplementedError`. There is a default value in the line just above this unpacking of left and right currents, so the code will not crash but just output a zero difference (also not ideal). I guess a small note could be helpful there

almaeder: Still, in the distributed case, are there no contact currents anymore?

vetschn: I removed that stuff completely, contact current output parameter gone as well. Probably not the best decision on second thought. I thought having this unified in one single device current buffer made sense, which i still think is the case for the non-distributed code.

The non-Meir-Wingreen device current computation is still supported in distributed mode. We could have the serinv backend also return the boundary currents directly using the Meir-Wingreen formula and stack that together with the regular device current for now. Any ideas @almaeder?

almaeder: Hmm, maybe have serinv only have the contact Meir Wingreen and keep the non-Meir Wingreen current separately?
Then the observables between dist and non-dist do not differ so much.

I believe some formulation of Meir Wingreen with the reduced system should be possible later.

vetschn: Okay so my proposition then:
- The `RGFDist` gets a `compute_current` parameter
- It returns only the Meir-Wingreen contact current at both ends, the rest of the buffer is filled with `NaN`
- We check at some later point how we can implement the device current formula directly in the solver (#271 )

almaeder: Makes sense to me 👍

src/quatrex/core/config.py
almaeder: This could be part of #231

vetschn: I agree that #231 might be a better place to fix this

almaeder (approved):
great workss!

PR #269, opened 2026-03-31
Support overlap matrices in NEGF computations

### LINKED ISSUE #226: Getting Blocks of sparse matrices
At multiple points in quatrex we want to get dense blocks from a sparse matrix. This code is repeated at several points:

```python
def get_block(
    coo: sparse.coo_matrix, block_sizes: NDArray, index: tuple
) -> NDArray:
    """Gets a block from a COO matrix."""
    block_offsets = np.hstack(([0], np.cumsum(block_sizes)))
    row, col = index
    row = row + len(block_sizes) if row < 0 else row
    col = col + len(block_sizes) if col < 0 else col
    mask = (
        (block_offsets[row] <= coo.row)
        & (coo.row < block_offsets[row + 1])
        & (block_offsets[col] <= coo.col)
        & (coo.col < block_offsets[col + 1])
    )
    block = xp.zeros(
        (int(block_sizes[row]), int(block_sizes[col])), dtype=coo.dtype
    )
    block[
        coo.row[mask] - block_offsets[row],
        coo.col[mask] - block_offsets[col],
    ] = coo.data[mask]

    return block
```

We should find a way to unify this.

### LINKED ISSUE #259: Properly support overlap matrix in NEGF
During merging of the k-points branch, there were bugs introduced that make NEGF calculations with overlap matrix incorrect. The support is currently dropped, but no issue was opened.

Problems are specifically in the assembly of the system matrix and in the computation of observables.

### LINKED ISSUE #260: The default `orthogonal_basis = True` leads to unexpected behaviour.
Even when there is a overlap matrix in the input directory the code silently assumes `orthogonal_basis = True` and continues while ignoring this overlap matrix.

There should either no default value at all to force the user to make an a priori decision, or the default is `None` and for the time being we stop the code if there is an overlap matrix detected in the input directory. (A warning is not enough imo).

Related to this: #259

vetschn:
- ~~Dynamically set the orthogonal_basis flag of the ElectronSolver, depending on the presence of `overlap.mat`.~~

- Extended the ElectronSolver with two kind of clunky helper methods, to add an overlap matrix on the block-tridiagonal and also to add the potential to the system matrix in a block-by-block fashion. Using these during assembly.

- Added a NEGF example with overlap matrix and only phonon pseudo-scattering. The block-size is rather large, so this will be a bit slow unfortunately.


src/quatrex/electron/solver.py
```diff
@@ -336,11 +417,6 @@ def _compute_obc(self) -> None:
                 gamma_00.copy(), self.left_occupancies - 1
             )
         if comm.block.rank == comm.block.size - 1:
-            # Extract the overlap matrix blocks.
-            s_nn = 1j * self.eta_obc * self._get_block(self.overlap_sparray, (-1, -1))
-            s_nm = 1j * self.eta_obc * self._get_block(self.overlap_sparray, (-1, -2))
-            s_mn = 1j * self.eta_obc * self._get_block(self.overlap_sparray, (-2, -1))
-
```
almaeder, line 338: Do we still need the `_get_block` function?
It seems to be used still for the band edges, but it could be replaced by the same logic.

src/quatrex/electron/solver.py
```diff
@@ -23,6 +23,80 @@
 profiler = Profiler()
 
 
+def _btd_add(a: DSDBSparse, b: DSDBSparse) -> None:
+    """Adds b to a on the block-tridiagonal.
```
almaeder, line 23: Should this, together with `_btd_subtract`, be combined with the iadd/isub methods of the data structure?

src/quatrex/electron/solver.py
```diff
@@ -23,6 +23,80 @@
...
+        a_.blocks[i, j] += b_.blocks[i, j]
+        a_.blocks[j, i] += b_.blocks[j, i]
+
+
+def _btd_apply_potential(
+    a: DSDBSparse, overlap: DSDBSparse, potential: NDArray
+) -> None:
+    """Applies the potential to a on the block-tridiagonal.
+
+    This is an in-place operation, i.e. a is modified.
```
almaeder, line 54: Shouldn't be handled by the data structure.

The docstring should contain that it does ~`S @ V + V @ S`.

src/quatrex/electron/solver.py
```diff
@@ -398,14 +485,18 @@ def _assemble_system_matrix(self, sse_retarded: DSDBSparse) -> None:
...
+            _btd_add(self.system_matrix, self.overlap)
 
         scale_stack(
             self.system_matrix.data,
             self.local_energies + 1j * self.eta,
         )
-        self.system_matrix -= sparse.diags(self.potential, format="csr")
+
+        if self.orthogonal_basis:
+            self.system_matrix -= sparse.diags(self.potential, format="csr")
```
almaeder, line 436: This is an already existing issue: `__iadd__` with scipy sparse matrices is rather inefficient. 
An option would be to get the diagonal, add the potential, and fill the diagonal. Also, in the orthogonal case, scaling does only needs to happen on the diagonal and not on the whole buffer.

src/quatrex/electron/solver.py
vetschn: Hmm, not so sure. I would expect that `__iadd__` does something different than this. Maybe the `btd_*` functionality could go into `datastructures/routines.py`?

src/quatrex/electron/solver.py
vetschn: What do you mean by "shouldn't be handled by the datastructure"? Agreed that the docstring should include this though

almaeder: In contrast to the comment before about `_btd_add`.

src/quatrex/electron/solver.py
almaeder: The difference is that currently `__iadd__` works only for commensurable matrices when the B (A += B) is dsdbsparse.
The  `_btd` methods always work, but ignore elements of B, which are not included in the sparsity pattern of A.

Suggested this since in the case of B being scipy sparse, `__iadd__` has the behavior of `_btd` (I believe). 
```
self.data += csr[
    self.rows + self.global_block_offset,
    self.cols + self.global_block_offset,
]
```
will ignore elements.

almaeder: Conclusion:
For now, we will keep this as is. If there will be a data structure refactor, this should be considered.

Reasoning:
- Expected effect of `__iadd__` is to be discussed.
- `_btd` method assumes a certain block pattern while `__iadd__` handles any. For this use case, `__iadd__` for non-commensurable matrices would work, but currently, such an implementation would lead to performance issues.

vetschn: Referencing #222 here as well.

src/quatrex/electron/solver.py
vetschn: Do you mean we should unify it (#226)? In the ElectronSolver i just noticed there is also a `@staticmethod` implementing the same thing :skull:

almaeder: Not unify here, but remove the  `_get_block` method from the electron solver. 
Only used for the bandedges now, which also could be replaced by the same logic as in the OBC.

almaeder: Maybe we keep this for now, but remove it together with the DOS peak tracking option.
That option does not seem to be recommended.

src/quatrex/electron/solver.py
almaeder, line 534: Should this be called with the `overlap_sparray`?
Also, similar to the assembly, could we pass in the flag if the overlap is identity instead of an eye matrix?

Then we could also remove the creation of the eye overlap sparray.

src/quatrex/electron/solver.py
almaeder: Similarly, there is a `get_block` in the band edges.

src/quatrex/electron/solver.py
vetschn: `overlap_sparray` is gone

src/quatrex/electron/solver.py
vetschn: they are all gone

vetschn:
cscs-ci run

almaeder:
cscs-ci run

almaeder (approved):
Changes to correctly support the overlap matrix make sense.
For the GW model, we would still need to investigate how the behavior is.

PR #273, opened 2026-04-14
Reduce workflow runs

vetschn:
Before the workflows were triggered on pull request and on push to all branches. In all PRs the workflows were always run twice.

The idea here is to switch this to only run once per pull request into `main` and `dev`, the `push` trigger is removed for the linting and the tests.

### DIFF
--- a/.github/workflows/linting.yaml
+++ b/.github/workflows/linting.yaml
@@ -1,7 +1,12 @@
 # Copyright (c) 2024-2026 ETH Zurich and the authors of the quatrex package.
 name: Run Pre-Commit and Determine Code Quality
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  pull_request:
+    branches:
+      - dev
+      - main
+  workflow_dispatch:
 
 jobs:
   test-linting:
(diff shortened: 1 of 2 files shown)


.github/workflows/linting.yaml
```diff
@@ -1,7 +1,12 @@
 # Copyright (c) 2024-2026 ETH Zurich and the authors of the quatrex package.
 name: Run Pre-Commit and Determine Code Quality
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  pull_request:
+    branches:
+      - dev
```
alexnick83, line 8: Won't this have the effect that only PR's for merging to dev/main run these tests?

vetschn: Yep, but we can have it be for all pull requests, also ones targeting other branches. Might be a better option

vetschn: I'll actually make it so that it triggers on all pull requests and also on pushes to `main` and `dev`, these we can use to maybe have a little workflow status badge in the readme later

alexnick83 (approved):

Issue #232, opened 2026-02-20
Deal with import-time side effects in Pydantic models


vetschn:
Some of the nested pydantic models, (e.g. `comm: CommConfig = CommConfig()` in `ComputeConfig`) are actually instantiated at module import which has the nasty side effect, that the validators on those submodels are called. Some of the validators configure singletons (`qtx_comm.configure(...)`) (oops) and these validators are called at import-time even before any provided config is parsed (💀).

We need to move global configuration side effects out of the validators. Importing modules must not have process-wide side effects.

almaeder:
https://github.com/quatrex/quatrex/pull/276 fixed this

PR #276, opened 2026-04-14
Pydantic side effects

almaeder:
This PR addresses #232.
This is done by removing the model validators that have side effects and moving them to a separate function.
That function is then called in the CLI/tests.

### DIFF
--- a/src/quatrex/core/config.py
+++ b/src/quatrex/core/config.py
@@ -26,7 +26,7 @@ from pydantic import (
 from typing_extensions import Self
 
 from qttools import xp
-from qttools.comm import comm as qtx_comm
+from qttools.comm import comm
 from qttools.datastructures import DSDBCOO, DSDBSparse
 from qttools.profiling import Profiler
 
@@ -1175,8 +1175,8 @@ def parse_config(config_file: Path) -> QuatrexConfig:
     return QuatrexConfig(**config)
 
 
-def _configure_profiler(config: QuatrexConfig) -> None:
-    """Configures the profiler based on the given configuration.
+def _setup_profiler(config: QuatrexConfig) -> None:
+    """Sets up the profiler based on the given configuration.
 
     Parameters
     ----------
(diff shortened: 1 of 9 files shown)


src/quatrex/core/config.py
```diff
@@ -1252,3 +1165,123 @@ def parse_config(config_file: Path) -> QuatrexConfig:
...
+        1,
+    ]:
+        if qtx_comm.rank == 0:
+            warnings.warn(
+                "The CPU code will run sequentially which may impact performance.",
+                UserWarning,
+            )
+
+
+def configure_qtx(config: QuatrexConfig) -> None:
```
vetschn: Sorry, have to pick on the names again. I think something like `setup_simulation_environment` or `setup_simulation_context`/ `setup_context` would be better. I think the `_internal` methods should then also be renamed to `_setup_*`

vetschn (changes requested):
Just the name of the additional method (and everything to do with it, also tests), everything else looks 💯 👌 👌

tests/quatrex/config/test_parse_config.py
```diff
@@ -4,16 +4,37 @@
...
     example_path, dist = example
 
     # Distributed examples should not validate the config parsing,
     # as they may contain distributed-specific options that are not
     # supported by the single-rank parser.
     with pytest.raises(ValueError) if dist else nullcontext():
         quatrex_config = example_path / "quatrex_config.toml"
-        parse_config(quatrex_config)
+        config = parse_config(quatrex_config)
+        configure_qtx(config)
```
vetschn: An additional sanity test might be to assert that comm and profiler are not already configured before any config is read (as was previously the case, #232 )

vetschn (approved):
I'm happy with it now. If you agree, feel free to merge @almaeder

tests/quatrex/config/test_parse_config.py
almaeder: Fair point, but it should be allowed to configure/setup multiple times. I need to check for any unexpected side effects from configuring twice.
I am guessing pytest also will do that since the tests run in the same session, right?

almaeder: Potential side effects:
- Creation of the sub communicator twice leads to an issue.
- Profiler will open multiple files.

For the last point, we need to set tmp directory as well before.
Otherwise, I believe, it will create an empty file profiling file in the example folders.

almaeder:
@vetschn, agree to merge with the extra fix?

PR #277, opened 2026-04-14
Load the config only on rank 0 

vetschn:
Previously all ranks read the config file more or less simultaneously. Now only rank zero does this. The resulting dictionary is broadcast (pickled). Each rank then validates the config seperately.

This is better than the validated config being sent around, because of all the side-effects that happen during validation.


almaeder (approved):
Good work

PR #281, opened 2026-04-16
Change pytest fixtures to remove autouse parameter where unnecessary

vetschn:
The use of `autouse` caused around 500 duplicate test cases.

Additionally, marking tests to use fixtures is unnecessary, when they already explicitly request a fixture. Just a cosmetic change.


almaeder (approved):
Thank you for the fix and it looks cleaner without the mark.

PR #282, opened 2026-04-17
Fix some of the issues determined by CodeQL

vetschn:


src/qttools/datastructures/routines.py
```diff
@@ -405,44 +405,70 @@ def btd_sandwich(
     out_.blocks[-2, -1] += a_.blocks[-2, -1] @ b_.blocks[-2, -1] @ a_.blocks[-1, -2]
 
 
-class BlockMatrix(dict):
+class BlockMatrix:
```
alexnick83, line 408: Does this still work properly without inheriting from `dict`?

src/qttools/datastructures/routines.py
```diff
@@ -405,44 +405,70 @@ def btd_sandwich(
...
+    Any local block keys are stored in the blocks of the dsdbsparse,
+    while non-local block keys are stored in a separate dictionary.
+
+    Parameters
+    ----------
+    dsdbsparse : DSDBSparse
+        The underlying DSDBSparse matrix.
+    local_keys : set[tuple[int, int]]
+        The set of block keys that are local to the current rank.
+    origin : tuple[int, int]
```
alexnick83, line 420: The description is a bit confusing; at least I was confused about exactly what the origin is and had to look ahead at the code. The origin is the global index of the first local block, i.e., local [0, 0] index in the partition.

src/quatrex/core/config.py
```diff
@@ -930,6 +930,7 @@ class BandEdgeConfig(BaseModel):
     block_sections: PositiveInt = 1
 
     @field_validator("use_eigvalsh", mode="after")
```
alexnick83, line 932: Are these decorators needed? How did it work before? I am assuming that `field_validator` includes some kind of `classmethod`

alexnick83 (approved):
It looks good to me as long as no functionality is erroneously changed.

src/qttools/datastructures/routines.py
vetschn: Yes, all dictionary features that were used was getitem and setitem, which were and are overridden anyway.

src/qttools/datastructures/routines.py
vetschn: Agreed, changing this

src/quatrex/core/config.py
vetschn: The `field_validator` decorator is actually intended to be used together with a `classmethod` decorator (https://pydantic.dev/docs/validation/latest/concepts/validators/#field-validators), which was missing before.

I guess the choice of `field_validator` here is better than model validator, because no cross-checks between different variables are made here.

It worked before because `self` was just passed into the method as `cls` parameter, but nothing is actually done with that. I think the `classmethod` decorator might just be to be more explicit that this must not be able to access the actual model state, but only the specific field

Issue #278, opened 2026-04-15
MoS2 Symmetry Problem


almaeder:
For MoS2, the reference checks do not hold when symmetry is True.

We should fix this and test both with and without symmetry.

vetschn:
@almaeder Is this fixed?

almaeder:
Is fixed.

Issue #279, opened 2026-04-16
No docstring for "symmetric" setting


YongDaLi:
https://github.com/quatrex/quatrex/blob/8ce585d87a5f606b76475956b45710177e9b232c/src/quatrex/core/config.py#L74

The symmetric setting has no docstring. But it interacts with other config settings, such as `align_self_energy_to_complex_axes`. It also changes the memory layout of `g_retarded.data`, which messes up my functions to save intermediate SCBA data. To be honest, I'm not sure what this setting actually does. There should be a docstring explaining it and its effects on other settings.

It can probably be fixed in the same PR that fixes this issue: https://github.com/quatrex/quatrex/issues/278

PR #283, opened 2026-04-17
Address symmetry issue when not aligning with the complex axis

almaeder:
This PR should fix the issues #278 and #279.

There are problems with the symmetry when not aligning to the complex axis or both P and Sigma.
The Hilbert parts generally contain imaginary parts, and when adding lesser/greater, the result is not complex symmetrical.
If we align, then they are symmetric.

We can fix this issue by keeping only the non-lesser/greater parts stored in Pr and SigmaR, and adding lesser/greater when Pr/SigmaR are used during the system matrix assembly G/W. 

This leads to extra additions, but we can further optimize when we align and store Pr/SigmaR as real instead of complex.

### DIFF
--- a/examples/w90/mos2/gw-kpoints-symmetric/reference-outputs
+++ b/examples/w90/mos2/gw-kpoints-symmetric/reference-outputs
@@ -1,0 +1,1 @@
+../gw-kpoints/reference-outputs/
--- a/src/quatrex/core/scba.py
+++ b/src/quatrex/core/scba.py
@@ -119,11 +119,11 @@ class SCBAData:
         self.sigma_greater_prev = dsdbsparse_type.zeros_like(self.g_lesser)
         self.sigma_greater = dsdbsparse_type.zeros_like(self.g_lesser)
 
-        self.sigma_retarded_prev = dsdbsparse_type.zeros_like(self.g_lesser)
-        self.sigma_retarded = dsdbsparse_type.zeros_like(self.g_lesser)
+        self.sigma_retarded_hermitian_prev = dsdbsparse_type.zeros_like(self.g_lesser)
+        self.sigma_retarded_hermitian = dsdbsparse_type.zeros_like(self.g_lesser)
         if config.scba.symmetric:
-            self.sigma_retarded.symmetry_op = lambda a: a.conj()
-            self.sigma_retarded_prev.symmetry_op = lambda a: a.conj()
+            self.sigma_retarded_hermitian.symmetry_op = lambda a: a.conj()
+            self.sigma_retarded_hermitian_prev.symmetry_op = lambda a: a.conj()
 
         if config.scba.coulomb_screening:
             # NOTE: The polarization has the same sparsity pattern as
@@ -350,9 +350,11 @@ class SCBA:
         """Stash the current into the previous self-energy buffers."""
         self.data.sigma_lesser_prev.data[:] = self.data.sigma_lesser.data
         self.data.sigma_greater_prev.data[:] = self.data.sigma_greater.data
-        self.data.sigma_retarded_prev.data[:] = self.data.sigma_retarded.data
+        self.data.sigma_retarded_hermitian_prev.data[:] = (
+            self.data.sigma_retarded_hermitian.data
+        )
 
-        self.data.sigma_retarded.data[:] = 0.0
+        self.data.sigma_retarded_hermitian.data[:] = 0.0
         self.data.sigma_lesser.data[:] = 0.0
         self.data.sigma_greater.data[:] = 0.0
 
@@ -365,14 +367,14 @@ class SCBA:
             # Make the self-energy Hermitian
             # This is done before adding the skew hermitian part coming
             # from the lesser and greater self-energies
-            self.data.sigma_retarded.symmetrize(xp.add)
+            self.data.sigma_retarded_hermitian.symmetrize(xp.add)
 
         if self.config.scba.align_self_energy_to_complex_axes:
             self.data.sigma_lesser._data.real = 0
             self.data.sigma_greater._data.real = 0
             # Make sure that the imaginary part comes only from
             # sigma_greater - sigma_lesser.
-            self.data.sigma_retarded._data.imag = 0
+            self.data.sigma_retarded_hermitian._data.imag = 0
 
     @profiler.profile(label="SCBA: Update Sigma", level="default", comm=comm)
     def _update_sigma(self) -> None:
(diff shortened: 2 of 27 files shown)


vetschn:
I was trying to backtrack where this assumption of "axis-alignment" actually comes from and i'm also starting to believe that this is an unfounded approximation to make. It is however on the reference `"printegral"` branch in the old quatrex (sometimes commented out, it is hard to know what was actually used):

```python
#SigL[ie] = 1j * np.imag(SigL[ie])
#SigG[ie] = 1j * np.imag(SigG[ie])

SigL[ie] = (SigL[ie] - SigL[ie].T.conj()) / 2
SigG[ie] = (SigG[ie] - SigG[ie].T.conj()) / 2
#SigR[ie] = np.real(SigR[ie]) + 1j * np.imag(SigG[ie] - SigL[ie]) / 2
SigR[ie] = np.real(SigR[ie]) + (SigG[ie] - SigL[ie]) / 2
#SigR[ie] = (SigR[ie] + SigR[ie].T) / 2
```

Instead of taking just real and imaginary parts like in that snippet, it would be better to consider `sigma_retarded` as the sum of a Hermitian and skew-Hermitian part (any matrix can be decomposed into this and i think it is analogous to real/imaginary part of complex scalars)

I was also wrong that Kramers-Kronig (Hilbert transform) only works between two real functions. It should work relating the Hermitian part of a matrix to the skew-Hermitian part.

So then the skew-Hermitian part of `sigma_retarded` should come from `0.5 * (sigma_greater - sigma_lesser)` and the Hermitian part should come from this via Kramers-Kronig. I don't think the full `sigma_retarded` actually ever really has general symmetry :grimacing:

Clearly not a big problem, we can just save the Hermitian part of `sigma_retarded` separately and assemble it on-the-fly like you already did anyways, @almaeder.

---

Some other stuff to quickly note here on the symmetry (just to have it written down):
- Retarded surface Green's functions have no symmetry properties (and neither the boundary self-energies)
- Lesser surface Green's function (Lyapunov) and self-energies are skew-Hermitian
- $A^{-1}X^{\lessgtr}A^{-\dagger}$ is skew-Hermitian for any non-symmetric complex $A$ and a skew-Hermitian $X^{\lessgtr}$ (which is consistent for system solve)
- $X \odot (i \cdot Y)$ is skew-Hermitian for skew-Hermitian $X$ and $Y$ (which is consistent for correlate)

examples/w90/mos2/gw-kpoints-symmetric/reference-outputs
vetschn, line 1: Was the idea with having this as a relative symlink to mainly save data on the git LFS (that would be handled automatically since hashes would match) or something else? Maybe more clear with the symlink that this is indeed the same data.

src/quatrex/core/scba.py
```diff
@@ -174,8 +174,8 @@ def __init__(self, config: QuatrexConfig, electron_energies: NDArray) -> None:
         self.sigma_retarded_prev = dsdbsparse_type.zeros_like(self.g_lesser)
         self.sigma_retarded = dsdbsparse_type.zeros_like(self.g_lesser)
         if config.scba.symmetric:
-            self.sigma_retarded.symmetry_op = lambda a: a
-            self.sigma_retarded_prev.symmetry_op = lambda a: a
+            self.sigma_retarded.symmetry_op = lambda a: a.conj()
+            self.sigma_retarded_prev.symmetry_op = lambda a: a.conj()
```
vetschn: I think here we could be a bit more explicit with the name, that this is `sigma_retarded_hermitian`.

src/quatrex/core/scba.py
```diff
@@ -759,10 +755,10 @@ def run(self) -> None:
...
             self._symmetrize_sigma()
 
-            # Add the anti-Hermitian part to the retarded self-energy.
-            self.data.sigma_retarded.data += 0.5 * (
-                self.data.sigma_greater.data - self.data.sigma_lesser.data
-            )
+            # # Add the anti-Hermitian part to the retarded self-energy.
+            # self.data.sigma_retarded.data += 0.5 * (
+            #     self.data.sigma_greater.data - self.data.sigma_lesser.data
+            # )
```
vetschn: Remove

src/quatrex/electron/solver.py
```diff
@@ -401,7 +456,7 @@ def _assemble_system_matrix(self, sse_retarded: DSDBSparse) -> None:
             self.local_energies + 1j * self.eta,
         )
         self.system_matrix -= sparse.diags(self.potential, format="csr")
-        _btd_subtract(self.system_matrix, sse_retarded)
+        _btd_subtract_sigma(self.system_matrix, sse_retarded, sse_lesser, sse_greater)
         _btd_subtract(self.system_matrix, self.hamiltonian)
```
vetschn: Might as well fuse these two into one function to avoid stepping through twice, no?

tests/quatrex/reference/test_reference.py
```diff
@@ -43,7 +43,7 @@ def test_single_rank(
             reference.shape == test.shape
         ), f"Shape mismatch for '{output_file.name}': {reference.shape} vs {test.shape}"
         assert np.allclose(
-            reference, test, rtol=1e-5, atol=1e-6
+            reference, test, rtol=1e-3, atol=1e-4
```
vetschn: ☠️

vetschn (approved):
I think it's fine to remove the retarded self-energy density quantity, these have been debug quantities anyway. If we end up deciding that we need this anyway, we can just reconstruct again from Hermitian / skew-Hermitian parts.

Rest looks good, kind of surprised how few changes this needs.

examples/w90/mos2/gw-kpoints-symmetric/reference-outputs
almaeder: To show that we test against the same data.

tests/quatrex/reference/test_reference.py
almaeder: 💀

src/quatrex/electron/solver.py
```diff
@@ -93,31 +93,63 @@ def _btd_apply_potential(
...
 
-def _btd_subtract(a: DSDBSparse, b: DSDBSparse) -> None:
-    """Subtracts b from a on the block-tridiagonal.
+def _btd_assemble(
+    system_matrix: DSDBSparse,
+    hamiltonian: DSDBSparse,
+    sse_lesser: DSDBSparse,
+    sse_greater: DSDBSparse,
+    sse_retarded_hermitian: DSDBSparse,
+) -> None:
```
vetschn: Before the method was quite a general helper method, but i would now put this into the ElectronSolver, i think you don't even have to treat this as a static method then, just operate on self.system_matrix

src/quatrex/coulomb_screening/solver.py
```diff
@@ -40,6 +40,53 @@ def _compute_sparsity_pattern(
     )
 
 
+def _btd_assemble_polarization(
+    p_retarded: DSDBSparse,
+    p_lesser: DSDBSparse,
+    p_greater: DSDBSparse,
+    p_retarded_hermitian: DSDBSparse,
```
vetschn: Like in the ElectronSolver, this probably also goes directly in the solver class, or at least rename the `p_retarded` parameter as `out`

vetschn (commented):
Just a quick thing about how this is structured, looks good otherwise 👌

vetschn (approved):
LGTM! Feel free to merge this in, if the tests pass

vetschn:
cscs-ci run

PR #285, opened 2026-04-28
Fix pseudo phonons

almaeder:
Previously, the shifting was not done correctly. The rolling and slicing did lead to unexpected results.
This was fixed by not rolling anymore, but by explicit slicing.

Note: It could be better to only take the valid part, but this follows OMEN/Matlab.

### DIFF
--- a/src/quatrex/electron/sse_phonon.py
+++ b/src/quatrex/electron/sse_phonon.py
@@ -40,34 +40,10 @@ class SigmaPhonon(ScatteringSelfEnergy):
                 self.phonon_energy, config.phonon.temperature
             )
 
-            # energy + hbar * omega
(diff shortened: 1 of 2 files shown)


src/quatrex/electron/sse_phonon.py
awinka, line 85: Do we need `sigma_retarded` in the `out` buffer? Currently we don't write anything to it

awinka (approved):
LGTM

vetschn (approved):
_LGTM_

vetschn:
The reference test data indeed does not have to change, which is a bit surprising. The energy grids we have in the example setups together with the phonon-frequency we set, does not actually cause a shift.

PR #286, opened 2026-04-28
Change the handling of the singularity in the Hilbert transformation.

awinka:
Instead of using a off-set `eta` to remove the singularity in the "Hilbert kernel", we set the singularity to zero. This removes the ambiguity in the off-set. 

Note: This is how it is done in the Matlab code.


almaeder:
It was discussed to use `2 * ne` instead of `2 * ne - 1`.
This change seems to be done here, except for the GW self-energy without the correction part, right?

awinka:
> It was discussed to use `2 * ne` instead of `2 * ne - 1`. This change seems to be done here, except for the GW self-energy without the correction part, right?

Ah, no. It currently uses the minimal padding `2 * ne - 1` everywhere.

almaeder:
Then we should stick with minimal padding for now and only change it if there is a performance reason.

tests/quatrex/electron/test_sse_coulomb_screening.py
```diff
@@ -8,12 +8,15 @@
...
-    hilbert_kernel = 1 / xp.concatenate([-energy_differences[::-1], energy_differences])
+    energy_differences = energies - energies[0]
+    # Remove the singularity by setting the energy difference to inf at the singularity.
+    energy_differences[0] = xp.inf
+    hilbert_kernel = 1 / xp.concatenate(
+        [-energy_differences[-1:0:-1], energy_differences]
+    )
     result = _naive_convolve(a_full, hilbert_kernel)
-    return result[2 * ne : 3 * ne]
+    return result[2 * ne - 1 : 3 * ne - 1]
```
vetschn, line 27: Why are the indices here and in the `test_polarization.py` different? Maybe just a quick comment would clear this up

vetschn (approved):
Only a small note, rest looks good 👍

almaeder (approved):
👍

tests/quatrex/electron/test_sse_coulomb_screening.py
awinka: Before the `hilbert_kernel` was size `2 * ne`, the "positive" and "negative" part of the kernel had `ne` points each. Now that we don't use a off-set, the kernel is one element smaller (`2 * ne - 1`) since we don't want to double count the 0 energy.

vetschn:
cscs-ci run

almaeder:
cscs-ci run

PR #287, opened 2026-04-29
Periodic W blocks

### LINKED ISSUE #97: Correct the assembled boundary blocks such that the multiplication is…
<a href="https://github.com/awinka"></img></a> **Issue by [awinka](https://github.com/awinka)**
_Wednesday Dec 11, 2024 at 10:26 GMT_
_Originally opened as https://github.com/vincent-maillou/quatrex/pull/15_

----

… handled properly. The assembled boundary blocks for the OBC calculations should not be taken directly from the system matrix, as that introduces blocks from further into the device of the polarizability in the OBC calculation. Instead the boundary blocks should be constructed from a specialized matrix multiplication.

----

_**[awinka](https://github.com/awinka)** included the following code: https://github.com/vincent-maillou/quatrex/pull/15/commits_

almaeder:
It was decided that the assumption should be that the OBC and spillover correction for W is done on fully periodic blocks.

This PR implements this.
The old assumptions could be valid, but there are still detailed investigations missing (for both approaches).
Limited changes would need to be made to roll back, since further code reuse is done now.


almaeder:
@vetschn we should discuss the changes first before updating the examples test data.

src/quatrex/core/utils.py
```diff
@@ -205,8 +209,86 @@ def get_periodic_superblocks(
...
+        a_ji=a_ji,
+        a_ii=a_ii,
+        a_ij=a_ij,
+        block_sections=block_sections,
+    )
+
+    for i in range(repetitions):
+        a_ii_out[
+            ...,
+            i * a_ii.shape[-1] : (i + 1) * a_ii.shape[-1],
```
vetschn: save `a_ii.shape[-1]` into a variable at the top

src/quatrex/core/utils.py
```diff
@@ -205,8 +209,86 @@ def get_periodic_superblocks(
...
+            i * a_ii.shape[-1] : (i + 1) * a_ii.shape[-1],
+        ] = a_ii_tmp
+
+    for i in range(repetitions - 1):
```
vetschn: Maybe leave a note/illustration that this can potentially neglect some blocks

src/quatrex/coulomb_screening/solver.py
```diff
@@ -156,6 +192,17 @@ def __init__(
         self.obc_blocks = OBCBlocks(num_blocks=self.system_matrix.num_local_blocks)
 
         self.block_sections = config.coulomb_screening.obc.block_sections
```
vetschn, line 175: Instead of having separate parameters for electrons and Coulomb Screening, we should just have a `NCpSC`-type parameter. So `num_cells_per_block` or something (`None` if constructed from unit cell) that should go in the `device` section of the config.

src/quatrex/coulomb_screening/solver.py
```diff
@@ -176,149 +223,262 @@ def _set_block_sizes(self, block_sizes: NDArray) -> None:
         self.l_lesser.block_sizes = block_sizes
         self.l_greater.block_sizes = block_sizes
 
-    @profiler.profile(label="CoulombScreeningSolver: OBC", level="default", comm=comm)
-    def _compute_obc(self) -> None:
-        """Computes open boundary conditions."""
-        if comm.block.rank == 0:
+    def _contact_obc(
```
vetschn:
```suggestion
    def _compute_contact_obc(
```

src/quatrex/coulomb_screening/solver.py
```diff
@@ -176,149 +223,262 @@ def _set_block_sizes(self, block_sizes: NDArray) -> None:
...
+        p_lesser : DSDBSparse
+            The lesser polarization.
+        p_greater : DSDBSparse
+            The greater polarization.
+        p_retarded : DSDBSparse
+            The retarded polarization.
+        diagonal_inds : tuple
+            The indices of the diagonal blocks corresponding to the contact.
+        upper_inds : tuple
+            The indices of the upper off-diagonal blocks corresponding to the contact.
```
vetschn, line 233: Maybe there is a way of passing only two indices? So layer index and the index of the layer towards the device? Not sure that this works on both sides easily

src/quatrex/coulomb_screening/solver.py
```diff
@@ -176,149 +223,262 @@ def _set_block_sizes(self, block_sizes: NDArray) -> None:
...
+                a_ji=_order_block(self.coulomb_matrix.blocks[*upper_inds[::-1]], order),
+                a_ii=_order_block(self.coulomb_matrix.blocks[*diagonal_inds], order),
+                a_ij=_order_block(self.coulomb_matrix.blocks[*upper_inds], order),
+                block_sections=self.small_block_sections,
+                repetitions=self.num_connected_blocks,
+            )
 
-                m_10_x_00 = m_10 @ x_00
-                self.obc_blocks.retarded[0] = m_10_x_00 @ m_01
+            m_00 = (
```
vetschn, line 277: Just on-liner note here that this is related for the spillover stuff

src/quatrex/coulomb_screening/solver.py
```diff
@@ -176,149 +223,262 @@ def _set_block_sizes(self, block_sizes: NDArray) -> None:
...
 
-                self.obc_blocks.lesser[0] = m_10 @ w_00_lesser @ m_10.conj().swapaxes(
-                    -1, -2
-                ) - (a_00_lesser - a_00_lesser.conj().swapaxes(-1, -2))
+            w_00, *__ = self.lyapunov((b_00, q_00), "W: " + contact)
 
-                self.obc_blocks.greater[0] = m_10 @ w_00_greater @ m_10.conj().swapaxes(
-                    -1, -2
-                ) - (a_00_greater - a_00_greater.conj().swapaxes(-1, -2))
+            m_w_m = m_10 @ w_00 @ m_10.conj().swapaxes(-1, -2)
```
vetschn, line 357: Maybe unpack here instead of `[0]` and `[1]`

src/quatrex/coulomb_screening/solver.py
```diff
@@ -176,149 +223,262 @@ def _set_block_sizes(self, block_sizes: NDArray) -> None:
...
-                comm=comm.stack,
-            ):
+            obc_greater = m_w_m[1] - (
+                a_00_greater - a_00_greater.conj().swapaxes(-1, -2)
+            )
 
-                n = self.system_matrix.num_local_blocks - 1
-                m = n - 1
+        return (
+            _order_block(obc_retarded, order),
```
vetschn: This will work for reverse, but if the intent is to support general permutations i don't think this works immediately.

src/quatrex/electron/solver.py
```diff
@@ -341,8 +341,8 @@ def _compute_obc(self) -> None:
 
             m_mn, m_nn, m_nm = get_periodic_superblocks(
                 # Twist it, flip it, ...
-                a_ii=xp.flip(self.system_matrix.blocks[n, n], axis=(-2, -1)),
                 a_ji=xp.flip(self.system_matrix.blocks[m, n], axis=(-2, -1)),
+                a_ii=xp.flip(self.system_matrix.blocks[n, n], axis=(-2, -1)),
                 a_ij=xp.flip(self.system_matrix.blocks[n, m], axis=(-2, -1)),
```
vetschn: Better to use the `_order_block` method here as well for consistency

src/quatrex/coulomb_screening/solver.py
almaeder: I opened #289 since it is a bit much to add here.

almaeder:
With this change, quatrex should reproduce the original MATLAB code one-to-one.

src/qttools/toeplitz/toeplitz.py
```diff
@@ -0,0 +1,212 @@
...
+
+    The periodic superblock structure will repeat the left- and
+    upper-most subblocks of the input block layer.
+
+    This does the following:
+    | aji | aii | aij |
+    | a b | c d | e 0 |
+    | 0 h | i j | k l |
+    then the periodic layer would be
+    | a i c d e |
```
vetschn: Should it be `a b c d e` here? Why are `aji`/`aii` there as well? Not sure that i understand the docstring (but it is already better than before)

src/qttools/toeplitz/toeplitz.py
```diff
@@ -0,0 +1,212 @@
...
+    leading to temporary superblocks
+    | aji | aii | aij |
+    | a i | c d | e 0 |
+    | 0 a | i c | d e |
+    if we want to double, then repeating only the periodic layer would give us
+    || 0 0 | a i || c d | e 0 || 0 0 | 0 0 ||
+    || 0 0 | 0 a || i c | d e || 0 0 | 0 0 ||
+    || 0 0 | 0 0 || a i | c d || e 0 | 0 0 ||
+    || 0 0 | 0 0 || 0 a | i c || d e | 0 0 ||
+
```
vetschn, line 155: Same here, i do not understand the notation, why are `aji` entries as well?

src/quatrex/core/utils.py
```diff
@@ -1,31 +1,6 @@
...
-    # if matrix.distribution_state != "stack":
-    #     raise ValueError("Matrix must be in stack distribution")
-
-    # for i in range(len(matrix.block_sizes) - 2):
-    #     matrix.blocks[i + 1, i + 1] = matrix.blocks[0, 0]
-    #     matrix.blocks[i + 1, i + 2] = matrix.blocks[0, 1]
-    #     matrix.blocks[i + 2, i + 1] = matrix.blocks[1, 0]
-
-    # matrix.blocks[-1, -1] = matrix.blocks[0, 0]
-    # matrix.blocks[-1, -2] = matrix.blocks[1, 0]
```
vetschn, line 4: 🥇

src/quatrex/coulomb_screening/solver.py
```diff
@@ -176,149 +189,275 @@ def _set_block_sizes(self, block_sizes: NDArray) -> None:
...
+            m_01 = -v_00 @ p_retarded_01 - v_01 @ p_retarded_00
+            m_10 = -v_10 @ p_retarded_00 - v_00 @ p_retarded_10
+
+        with profiler.profile_range(
+            label=f"CoulombScreeningSolver: OBCR {contact}",
+            level="default",
+            comm=comm.stack,
+        ):
+
+            x_00, *__ = self.obc((m_00, m_01, m_10), contact="W: " + contact)
```
vetschn, line 292: Why the `W: ` prefix?

src/quatrex/coulomb_screening/solver.py
```diff
@@ -176,149 +189,275 @@ def _set_block_sizes(self, block_sizes: NDArray) -> None:
...
-                    )
-                    @ x_nn.conj().swapaxes(-1, -2)
-                )
+        The result of this method is that the `obc_blocks` attribute of the solver is filled.
+
+        NOTE: The polarizations are passed as arguments and not the system matrix.
+        This is because not the blocks of the system matrix are used, but
+        fully periodic superblocks are assembled with the polarizations and the Coulomb matrix.
+        In the case of no subdivision, the system matrix blocks could be used directly.
+
```
vetschn, line 392: PEP8 formatting (72 characters per line)

```suggestion
        """Computes open boundary conditions (OBC).

        Both the OBC for retarded and lesser/greater components are
        computed, as the former is needed for the latter. This done for
        all the contacts of the system, which are currently assumed to
        be only the left and right boundaries.

        The result of this method is that the `obc_blocks` attribute of
        the solver is filled.

        NOTE: The polarizations are passed as arguments and not the
        system matrix. This is because not the blocks of the system
        matrix are used, but fully periodic superblocks are assembled
        with the polarizations and the Coulomb matrix. In the case of no
        subdivision, the system matrix blocks could be used directly.

```

src/qttools/toeplitz/toeplitz.py
```diff
@@ -0,0 +1,212 @@
...
+    """Homogenizes a matrix in stack distribution.
+
+    Parameters
+    ----------
+    matrix : DSDBSparse
+        The matrix to homogenize.
+    """
+
+    raise NotImplementedError()
+
```
vetschn, line 18: Deprecate this in a separate PR?

src/quatrex/electron/solver.py
```diff
@@ -316,101 +317,110 @@ def _update_fermi_levels(
...
+        else:
+            # Extract the overlap matrix blocks.
+            s_10 = 1j * self.eta_obc * self.overlap.blocks[*upper_inds[::-1]]
+            s_00 = 1j * self.eta_obc * self.overlap.blocks[*diagonal_inds]
+            s_01 = 1j * self.eta_obc * self.overlap.blocks[*upper_inds]
+
+        # TODO: use residuals to filter "bad" energies
+        g_00, *__ = self.obc(
+            (m_00 + s_00, m_01 + s_01, m_10 + s_10),
+            contact="G: " + contact,
```
vetschn, line 279: Also here, why the prefix?

vetschn (commented):
Looks pretty good, but some small questions came up while looking through

src/qttools/toeplitz/toeplitz.py
almaeder: `aij` / `aii` as in the name of the block. Maybe it would be clearer as:
```
      aji   aii   aij 
    | a b | c d | e 0 |
    | 0 h | i j | k l |
```
`a b c d e` is intended for the names of the smaller subsections. Having something like `aji_00 aji_01` would be possible, but I think it would not be clearer.

src/quatrex/coulomb_screening/solver.py
almaeder: To improve the warnings when the recursion error is not "good". 
Otherwise, they alias between G and W, and I am never sure which is not working well:)

src/qttools/toeplitz/toeplitz.py
almaeder: I would not deprecate it if it is really needed for flatband simulations.
Sometimes, a flatband simulation will be needed to check the correctness of other models.

src/quatrex/electron/solver.py
almaeder: See other reply.

almaeder:
cscs-ci run

vetschn (approved):
Nice, i think might be even clearer now that the full polarization is explicitly assembled in the body of `solve()`

PR #292, opened 2026-05-05
Dielectric environment

awinka:
Functionality to include a dielectric environment that modifies the screened Coulomb interaction.
[You can add a longer description of what the changes do]


examples/w90/hBN/quatrex_config_hBN.toml
awinka, line 1: I would suggest that you can create a new folders for the dielectric environment simulations. For example one for hBN and one for rpa.

It is also better if you run your simulations elsewhere, for example on a scratch (I can show you how in a meeting).

examples/w90/carbon-nanotube/gw-unit-cell/plotting.ipynb
awinka, line 1: We usually don't push postprocessing/plotting scripts to the remote, especially if they have outputs, since they change a lot.

It can be a good idea to track changes to important notebooks, but you can then clear the outputs before commiting so you only track the contents of the cells.

src/quatrex/cli/main.py
awinka, line 1: This is can be a nice option. You can compute the environment screened Coulomb interaction and then save it to disk for later re-use. in for example a scba calculation. I also like that it have a separate [environment] section in the config.

However, I could not find any functionality for loading the saved environment screened Coulomb interactions. So if this is a feature we want to keep, we also should add that

src/quatrex/core/config.py
```diff
@@ -594,69 +529,104 @@ class CoulombScreeningConfig(BaseModel):
...
+    
+    If not set, the retarded polarization is computed only from
+    the lesser and greater parts by $\frac{\mathbf{P}^> - \mathbf{P}^<}{2}$.
+
+    """
 
-    Possible values are `"self-energy"`, `"polarization"`, and `"both"`.
+    polarization_method: Literal["negf", "rpa"] = "negf"
+    dielectric_environment: bool = False
+    dielectric_method: Literal["rpa"] | None = None
```
awinka: Here I think we also can add an option to calculate the polarization with Green's functions also.

src/quatrex/core/config.py
awinka, line 615: These changes should go into the new EnvironmentConfig. We don't want to calculate the scba Polarization with the RPA, only the environment Polarization. The reason is that the RPA is more suitable for a equilibrium calculation, which we assume in the environment.

Many of these setting are also for the RPA calculation, so you could create a separate RPAConfig also.

src/quatrex/core/scba.py
```diff
@@ -452,24 +573,62 @@ def _compute_coulomb_screening_interaction(self):
...
+                self.data.p_retarded,
+                self.data.p_lesser,
+                self.data.p_greater,
+            )
+        else:
+            self.p_coulomb_screening.compute(
+                self.data.g_lesser,
+                self.data.g_greater,
+                out=(self.data.p_lesser, self.data.p_greater, self.data.p_retarded),
+            )
```
awinka: In the scba, we don't want to calculate the polarization with the RPA, so you can remove this

src/quatrex/core/scba.py
```diff
@@ -158,6 +218,19 @@ def __init__(self, config: QuatrexConfig, electron_energies: NDArray) -> None:
...
+            ):
+                self.w_environment_retarded = dsdbsparse_type.from_sparray(
+                    self.sparsity_pattern.astype(xp.complex128),
+                    block_sizes=block_sizes,
+                    global_stack_shape=electron_energies.shape
+                    + tuple([k for k in kpoint_grid if k > 1]),
+                    symmetry=False,
+                )
+                self.w_environment_lesser = dsdbsparse_type.zeros_like(self.p_lesser)
+                self.w_environment_greater = dsdbsparse_type.zeros_like(self.p_greater)
```
awinka, line 233: Good! I think it is a good idea to create separate buffers for the environment screened Coulomb interactions.

I'm not sure that you also need to check if the `dielectric_method` have a valid value here. I think it is fine with only the `dielectric_environment` condition

src/quatrex/core/scba.py
```diff
@@ -347,6 +441,31 @@ def __init__(self, config: QuatrexConfig) -> None:
...
+            )
+            self.coulomb_screening_solver.l_lesser.free_data()
+            self.coulomb_screening_solver.l_greater.free_data()
+
+            self.rpa_coulomb_screening_bridge = EquilibriumRPAScreeningBridge(
+                self.config,
+                self.coulomb_screening_energies,
+                self.data.p_lesser,
+            )
+
```
awinka, line 474: Good! I think this structure is good

Since we also want to support the GW method to calculate the polarization, you could rename things to maybe `environment_coulomb_screening_bridge` instead of `rpa_coulomb_screening_bridge`

src/quatrex/core/scba.py
```diff
@@ -452,24 +573,62 @@ def _compute_coulomb_screening_interaction(self):
...
+            )
+            self.coulomb_screening_solver.solve(
+                self.data.p_lesser,
+                self.data.p_greater,
+                self.data.p_retarded,
+                out=(self.data.w_lesser, self.data.w_greater),
+                interaction_retarded=self.data.w_environment_retarded,
+                interaction_lesser=self.data.w_environment_lesser,
+                interaction_greater=self.data.w_environment_greater,
+            )
```
awinka: Good! It is very nice that the `rpa_coulomb_screening_bridge` have a cache, such that the environment screened interactions aren't recomputed.

One issue though is that these matrices are not the environment screened Coulomb interactions that we want. What you should solve are the equations in the `block_screening.py` module.

src/quatrex/core/scba.py
```diff
@@ -452,24 +573,62 @@ def _compute_coulomb_screening_interaction(self):
...
+                self.data.p_lesser,
+                self.data.p_greater,
+                self.data.p_retarded,
+                out=(self.data.w_lesser, self.data.w_greater),
+            )
 
         self._compute_coulomb_screening_observables()
 
+        if self.config.outputs.full_polarization:
+            self._write_full_polarization(self.iteration)
```
awinka: Why do you write the full polarization?

src/quatrex/core/scba.py
```diff
@@ -493,31 +652,82 @@ def _compute_coulomb_screening_interaction(self):
...
+            ("p_retarded", self.data.p_retarded),
+        ):
+            dense_local = matrix.to_dense()
+            dense = comm.stack.all_gather_v(
+                dense_local,
+                axis=0,
+                mask=matrix._stack_padding_mask,
+            )
+            if comm.rank == 0:
+                xp.save(output_dir / f"{name}_full_{iteration}.npy", dense)
```
awinka: I'm not sure this is needed

src/quatrex/coulomb_screening/dielectric_screening/equilibrium_screening.py
awinka, line 1: Nice! It seems that this class calculates the equilibrium polarization and screened Coulomb interaction. The results from this class is then used in the scba loop through the `negf_bridge.py` module.

It provides functionality for many different formats such as `grid` and `matrix`, and we probably don't want to support all of them. For example, we are only interested in the `matrix` format, where the polarization also are orbital resolved. I also don't think that we only want to compute quanties for a single (q, omega) point

src/quatrex/coulomb_screening/dielectric_screening/negf_bridge.py
awinka, line 1: Very nice script! It provides some very useful functionality for bridging the equilibrium quantities to scba.

src/quatrex/coulomb_screening/dielectric_screening/negf_bridge.py
```diff
@@ -0,0 +1,326 @@
...
+        )
+
+        for local_index in range(local_energy_count):
+            global_index = energy_offset + local_index
+            if w_retarded is not None:
+                w_retarded.stack[(local_index,)] = result.w_retarded_matrices[
+                    global_index
+                ]
+            w_lesser.stack[(local_index,)] = result.w_lesser_matrices[global_index]
+            w_greater.stack[(local_index,)] = result.w_greater_matrices[global_index]
```
awinka, line 208: Very cool! I'm not sure the datastructure supports this sort of setting, will have to doublecheck

src/quatrex/coulomb_screening/dielectric_screening/negf_bridge.py
```diff
@@ -0,0 +1,326 @@
...
+        local_energy_count = int(self.template.stack_section_sizes[self._stack_rank])
+        energy_offset = int(
+            np.sum(self.template.stack_section_sizes[: self._stack_rank])
+        )
+
+        for local_index in range(local_energy_count):
+            global_index = energy_offset + local_index
+            p_retarded.stack[(local_index,)] = result.p_retarded_matrices[global_index]
+            p_lesser.stack[(local_index,)] = result.p_lesser_matrices[global_index]
+            p_greater.stack[(local_index,)] = result.p_greater_matrices[global_index]
```
awinka: This will probably not be used and can be removed

src/quatrex/coulomb_screening/dielectric_screening/negf_bridge.py
```diff
@@ -0,0 +1,326 @@
...
+
+    @property
+    def _stack_rank(self) -> int:
+        return int(comm.stack.rank)
+
+    def _build_cached_result(self) -> EquilibriumRPABridgeResult:
+        result = None
+        if global_comm.rank == 0:
```
awinka, line 490: This is very inefficient. Now only rank 0 computes the results and broadcast it. Ideally the workload should be split over the ranks. However, for now it is probably fine since we are only interested in correctness

src/quatrex/coulomb_screening/dielectric_screening/negf_bridge.py
```diff
@@ -0,0 +1,326 @@
...
+                block_index = [0, 0, 0]
+                block_index[periodic_axis] = translation + max_translation
+                unit_cells[tuple(block_index)] = block
+
+            matrix_sparray, __, __ = _create_matrix_from_unit_cells(
+                self.config, unit_cells
+            )
+            matrices.append(matrix_sparray.astype(xp.complex128))
+
+        return matrices
```
awinka, line 799: Very cool method. Upscaling the unit cell quantities to the full device.

src/quatrex/coulomb_screening/dielectric_screening/polarization.py
awinka, line 1: What is the usefulness of this module? It seems to provide a similar interface to the PCoulombScreening class from regular quatrex, but using the RPA method instead of the normal GW method.

I can not see it being used anywhere in the program.

src/quatrex/coulomb_screening/dielectric_screening/rpa_compute.py
awinka, line 1: This script provides the functions for actually computing the rpa polarization. At a first glance it looks correct, but have to be tested.

Once again there is a  `_matrix` option for the solver, which I believe is the only solver we should support.

Currently it is support only for k-points/periodicity in the transport direction. For now this is fine, but in the future we want to make it more general.

src/quatrex/coulomb_screening/dielectric_screening/equilibrium_screening.py
```diff
@@ -0,0 +1,511 @@
...
+            polarization_value = polarization_grid[q_index, frequency_index]
+            if polarization_grid.ndim == 2:
+                dielectric_matrix = identity - polarization_value * coulomb_matrix
+            else:
+                dielectric_matrix = identity - coulomb_matrix @ polarization_value
+            dielectric_matrices[q_index, frequency_index] = dielectric_matrix
+            screened_interactions[q_index, frequency_index] = np.linalg.solve(
+                dielectric_matrix,
+                coulomb_matrix,
+            )
```
awinka, line 132: These for loops are gonna be super slow to compute the `dielectric_matrices` and `screened_interactions`. It should be straightforward to vectorize this to make it faster.

src/quatrex/coulomb_screening/dielectric_screening/equilibrium_screening.py
```diff
@@ -0,0 +1,511 @@
...
+    matrix = np.asarray(coulomb_matrix, dtype=np.complex128)
+    if matrix.ndim != 2 or matrix.shape[0] != matrix.shape[1]:
+        raise ValueError("coulomb_matrix must be a square matrix.")
+    identity = np.eye(matrix.shape[0], dtype=np.complex128)
+    polarization_array = np.asarray(polarization, dtype=np.complex128)
+    if polarization_array.ndim == 0:
+        return identity - np.complex128(polarization_array) * matrix
+    if polarization_array.shape != matrix.shape:
+        raise ValueError("matrix-valued polarization must match coulomb_matrix shape.")
+    return identity - matrix @ polarization_array
```
awinka: Will probably never use this. It also looks like it is never used.

src/quatrex/coulomb_screening/dielectric_screening/equilibrium_screening.py
```diff
@@ -0,0 +1,511 @@
...
+def compute_screened_coulomb_matrix(
+    coulomb_matrix: NDArray[np.complex128],
+    polarization: complex | NDArray[np.complex128],
+) -> tuple[NDArray[np.complex128], NDArray[np.complex128]]:
+    """Compute ``epsilon`` and ``W = epsilon^{-1} V``."""
+
+    dielectric_matrix = compute_dielectric_matrix(coulomb_matrix, polarization)
+    screened_interaction = np.linalg.solve(dielectric_matrix, coulomb_matrix)
+    return dielectric_matrix, screened_interaction
+
```
awinka: Same with this. We will always compute for many `q`-points and energies `omega` at the same time

src/quatrex/coulomb_screening/dielectric_screening/equilibrium_screening.py
```diff
@@ -0,0 +1,511 @@
...
+            inputs=inputs,
+            mesh=mesh,
+            chemical_potential=chemical_potential,
+            temperature=temperature,
+            q_index=q_index,
+            frequency_index=frequency_index,
+            periodic_axis=periodic_axis,
+            lattice_constant=lattice_constant,
+            broadening=broadening,
+        )
```
awinka: Same here. These are a bit unnecessary and can be removed.

src/quatrex/coulomb_screening/dielectric_screening/equilibrium_screening.py
```diff
@@ -0,0 +1,511 @@
...
+            coulomb_blocks=coulomb_blocks,
+            mesh=mesh,
+            chemical_potential=chemical_potential,
+            temperature=temperature,
+            periodic_axis=periodic_axis,
+            lattice_constant=lattice_constant,
+            broadening=broadening,
+            q_index=q_index,
+            frequency_index=frequency_index,
+        )
```
awinka: This only computes the  results at a single `q_index` and `frequency_index`, which is not very useful and will probably never be used. Can be removed

src/quatrex/coulomb_screening/dielectric_screening/negf_bridge.py
```diff
@@ -0,0 +1,326 @@
...
+        print("RPA cache: transforming greater W to transport matrices...", flush=True)
+        w_greater_matrices = self._build_transport_matrices(mesh, w_greater_qw)
+        return EquilibriumRPABridgeResult(
+            p_retarded_matrices=p_retarded_matrices,
+            p_lesser_matrices=p_lesser_matrices,
+            p_greater_matrices=p_greater_matrices,
+            w_retarded_matrices=w_retarded_matrices,
+            w_lesser_matrices=w_lesser_matrices,
+            w_greater_matrices=w_greater_matrices,
+        )
```
awinka, line 708: Calculations seems to be correct

src/quatrex/coulomb_screening/dielectric_screening/rpa_compute.py
```diff
@@ -0,0 +1,1041 @@
...
+        periodic_axis = infer_periodic_axis(translation_blocks)
+    if periodic_axis not in (0, 1, 2):
+        raise ValueError("periodic_axis must be 0, 1, or 2.")
+
+    first_block = next(iter(translation_blocks.values()))
+    norb = first_block.shape[0]
+    bloch_hamiltonian = np.zeros((k_points_array.size, norb, norb), dtype=np.complex128)
+    for translation, block in translation_blocks.items():
+        phase = np.exp(
+            1j * k_points_array * translation[periodic_axis] * lattice_constant
```
awinka, line 235: Make sure the `1j` or `-1j` convention is the same as in quatrex. It should not matter since the matrices are symmetric, but it is good for consistency.

src/quatrex/coulomb_screening/dielectric_screening/rpa_compute.py
```diff
@@ -0,0 +1,1041 @@
...
+            integrand
+            * k_weights_array[np.newaxis, :, np.newaxis, np.newaxis, np.newaxis]
+        )
+        polarization = integrand.sum(axis=(1, 2, 3), dtype=np.complex128)
+    else:
+        polarization = integrand.sum(axis=(1, 2, 3), dtype=np.complex128)
+        if normalize_k_sum and nk > 0:
+            polarization = polarization / nk
+
+    return polarization
```
awinka, line 619: This option without `_matrix` can be removed.

src/quatrex/coulomb_screening/dielectric_screening/rpa_compute.py
```diff
@@ -0,0 +1,1041 @@
...
+        return compute_rpa_polarization_from_bands(
+            band_structure,
+            mesh.q_points,
+            mesh.frequencies,
+            chemical_potential=chemical_potential,
+            temperature=temperature,
+            state_multiplicity=self.channels.multiplicity,
+            broadening=broadening,
+            frequency_axis=self.frequency_axis,
+        )
```
awinka, line 801: Non-orbital-resolved option can be removed

src/quatrex/coulomb_screening/dielectric_screening/rpa_compute.py
```diff
@@ -0,0 +1,1041 @@
...
+        return self.solve_from_translation_blocks(
+            translation_blocks=translation_blocks,
+            mesh=mesh,
+            chemical_potential=chemical_potential,
+            temperature=temperature,
+            periodic_axis=periodic_axis,
+            lattice_constant=lattice_constant,
+            broadening=broadening,
+        )
```
awinka, line 913: Non-orbital-resolved options can be removed.

src/quatrex/coulomb_screening/dielectric_screening/solver.py
awinka, line 1: This is similar to the `src/quatrex/coulomb_screening/dielectric_screening/polarization.py` module. It is currently not used anywhere.

It probably could be useful for a equilibrium GW calculation, which is work that is underway on another branch.

src/quatrex/coulomb_screening/block_screening.py
awinka, line 1: I can not see this module being used anywhere. I don't find it particularly useful either. Maybe it can be used as a reference for tests.

It is good however that it implements the correct equations for the environment screened Coulomb interaction. These equations are not implemented anywhere else.

src/quatrex/coulomb_screening/environment.py
awinka, line 1: This looks like the equation for the Coulomb matrix based on the paper `Theoretical determination of the effect of a screening gate on plasmon-induced superconductivity in twisted bilayer graphene`. It is not something that we really can use as is.

This file can probably be removed.

src/quatrex/coulomb_screening/environment_export.py
awinka, line 1: This is a nice idea (see my previous comment in the `quatrex_config.py` file), but functionality for loading the Polarization and screened interaction is not in place. So this is something we can keep on hold for now.

src/quatrex/coulomb_screening/solver.py
awinka, line 1: Very nice changes. It looks correct, but has to be tested further.

tests/quatrex/config/test_parse_rpa_screening_config.py
awinka, line 1: Nice tests.

tests/quatrex/coulomb_screening/test_environment.py
awinka, line 1: This test can be removed, since the `environment.py` module is not used.

src/quatrex/coulomb_screening/dielectric_screening/rpa_compute.py
```diff
@@ -0,0 +1,1041 @@
...
+    return blocks
+
+
+def resolve_unit_cell_matrix_path(
+    config: QuatrexConfig,
+    matrix_name: str,
+) -> Path:
+    """Resolve a unit-cell matrix path from the Quatrex input directory."""
+
+    return Path(config.input_dir) / f"{matrix_name}.mat"
```
awinka: This will read the inputs to the scba calculation, it should probably be something like `config.environment.input_dir` to use the environment matrices instead

awinka (commented):
Some very nice contributions!

You can have a look at the comments for the files. I've only read the code, I haven't run it yet, so the comments are based on that.

I think the structure for the code is good. There is a problem with the `negf_bridge.py`; it computes the equilibrium properties in the environment but not the environment screened Coulomb interaction (as are implemented in the script I sent you and in the `block_screening.py` module).

We should also try to add the functionality to calculate the environment polarization using the GW approximation in equilibrium, not only having the RPA option. Then we can have that as a fallback in-case we don't manage to make the RPA work

calmestelle:
SCBA now computes device polarization from NEGF Green's functions, the RPA bridge is only used for cached environment-screened interactions, full polarization output is disabled in the hBN config, local output/debug artifacts are ignored, and the bridge test import is fixed. Larger follow-ups remain for saved environment interaction loading, GW equilibrium fallback, and example reorganization

src/quatrex/core/scba.py
calmestelle: I think i included this while trying to debug the RPA polarization

examples/w90/carbon-nanotube/gw-unit-cell/plotting.ipynb
calmestelle: would you like me to remove them? I tried doing that, but I think they are considered as LFS, and I don't have access to GitLab to make any changes

awinka: It is fine for now. Down the line we want to remove them, but I don't think we have to worry about it now

Issue #274, opened 2026-04-14
Coulomb Matrix in point-charge approximation


vetschn:
An alternative to explicitly providing a Coulomb matrix to the code could be to construct a matrix from the atomic structure, assuming point charges

$$
v_{ij} = \frac{q}{4\pi \varepsilon_0 \varepsilon_r \vert \mathbf{r}_i - \mathbf{r}_j \vert}
$$

This can happen directly at runtime.

Issue #270, opened 2026-04-13
Improved convergence information CLI output


vetschn:
We should write more detailed/meaningful convergence information to `stdout`, such as the relative current convergence, convergence across the entire device, etc. This could maybe even be done using nicely formatted table using `typer`/`click`.

Issue #117, opened 2025-09-05
Maybe encapsulate properties relating to a device into a `Device` class?


almaeder:
<a href="https://github.com/vetschn"><img src="https://avatars.githubusercontent.com/u/63104434?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [vetschn](https://github.com/vetschn)**
_Friday May 23, 2025 at 09:40 GMT_
_Originally opened as https://github.com/vincent-maillou/quatrex/issues/65_

----

We have several quantities that are all related to the physical structure/device we are simulating. It may be worth bundling this together into an object. I'm thinking of things like:
- Hamiltonian matrix
- Overlap matrix
- Coulomb matrix
- Dynamical matrix
- Location of orbital centers
- Optical/thermal interaction tensors
- Location of contacts (and other things relating to contacts)
- Maybe even things concerning the electrostatics (gate/oxide/mesh)

This could be very neat but we would need to be pretty careful not to accidentally instantiate copies of this object.

Issue #114, opened 2025-09-05
Generalize Contact Implementation


almaeder:
<a href="https://github.com/vetschn"><img src="https://avatars.githubusercontent.com/u/63104434?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [vetschn](https://github.com/vetschn)**
_Friday May 23, 2025 at 08:45 GMT_
_Originally opened as https://github.com/vincent-maillou/quatrex/issues/62_

----

> Mostly to open the discussion: wouldn't it make more sense to call those source/drain or whatever applies appropriately?

_Originally posted by @alexnick83 in https://github.com/vincent-maillou/quatrex/pull/57#discussion_r2099526464_

I think this would definitely make sense. Generally, we need to rethink some things concerning how a user of the program could define contacts. Then i think having a number of Contact objects would be a reasonable approach. These could handle determining the band edges, the boundary self energies etc.

@maurodossena came up with a scheme already for the QTBM to select the contact orbitals from their real-space location. This map is then used to apply the contact self-energy in the correct places in the system matrix. We need to evaluate what the best approach is to accomplish something similar in NEGF.

almaeder:
It will be important that we work with orbital positions and not atom positions.

vetschn:
@almaeder Is this addressed?

PR #294, opened 2026-05-07
Electrostatics

### LINKED ISSUE #96: Poisson solver: first implementation
<a href="https://github.com/vincent-maillou"></img></a> **Issue by [vincent-maillou](https://github.com/vincent-maillou)**
_Wednesday Aug 21, 2024 at 09:32 GMT_
_Originally opened as https://github.com/vincent-maillou/quatrex/issues/11_

----

None

vetschn:
This pull contains many additions and some changes concerning electrostatics / self-consistent charge density calculations.

**Meshing**
- There is now a kind of CSG handling for creating FEM meshes from the config, similar to what OMEN was doing.
- The meshing is based on `gmsh`/`OCC` and supports periodic boundary conditions via multi-freedom constraints (for calculations including k-points in transverse directions) and definitions of box and cylindrical volumes (with settings for dopants and relative permittivities), as well as rectangular and cylindrical surfaces for Dirichlet boundary conditions.
- There is also a new `mesh` CLI command to mesh and visualize the geometry definition. This has to be called before a self-consistent simulation is started. It is also ensured that mesh and config are consistent during the self-consistent Schrödinger-Poisson (SCSP) simulation.

**Poisson solvers and potential mixing**
- There are two types of Poisson solvers, a direct one and one based on root-finding, combined with a model of the "density response" to potential changes. The root-finding scheme should be better in most cases (this is what OMEN used).
- The single-band density response models extend what OMEN was doing by also allowing density models of materials with dimensions different than 2D. These models involve computing Fermi integrals and their inverses, which is simple for 2D, where this has an analytic form. There are approximate schemes for computing these integrals/inverses for dimensions other than 2D, which is also implemented now. ***This addition is likely overkill, but should in principle more accurately model the density and ideally accelerate the convergence. This will have to be tested in some different systems.***
- I added two kinds of mixers to update the potential, an under-relaxation one and a simple DIIS implementation, because we wanted to test whether this accelerates convergens. ***We will also have to test this in different systems and should extend the under-relaxation to adaptive under-relaxation, which could then also be used to update the self-energy in SCBA***

**Self-Consistent Calculation**
- Implemented `set_potential`, `compute_charge_density`, and `_compute_excess_charge_densities` methods for both QTBM and SCBA solvers.
- These methods enable the calculation of excess charge densities from the DOS / spectral carrier densities and enable updating the potential for self-consistent calculations.

**Physical Constants and Units**
- ~~Changed the definition of the speed of light constant `c_0` to use nm/s instead of angstrom/s (`c_0` was there but not really used before). This changes the units of `epsilon_0` and `mu_0`, that are computed via the fine structure relation. This change is needed because meshing is done in nm (to compare agains OMEN).~~
- Updated the SCBA observable calculations to consistently account for spin by multiplying densities by 2. This will have to be parameterized down the line if we want to do SOC calculations.

**Remaining Tasks**
- [x] Include automatic Fermi level calculation.
- [x] Compute the enforced gate Dirichlet boundary conditions from metal work function / electron affinity
- [x] Since contact Fermi levels should be computed automatically, we need to define relative voltages between contacts.
- [x] Update package dependencies (`ase`/`skfem`/...)
- [ ] ~~Support meshing from unit cell~~ #322 
- [x] Potential restarting from file appears to double count (?)
- [x] GPU testing and implementation
- [x] Integration test
- [x] Better starting guess
- [ ] Unit tests
- [x] Update Docs build
- [x] Consume Electron affinity to correctly set gate Dirichlet boundary conditions


src/quatrex/electrostatics/mixer.py
```diff
@@ -0,0 +1,186 @@
...
+from qttools import NDArray
+
+
+class Mixer(ABC):
+    """Abstract base class for mixing schemes."""
+
+    @abstractmethod
+    def mix(self, previous_value: NDArray, incoming_value: NDArray) -> NDArray:
+        """Mixes the incoming value with the previous value."""
+        ...
```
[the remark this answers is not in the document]
vetschn: In abstract methods `Ellipsis` makes more sense than raising exceptions and `pass` statements. `Ellipsis` is more like "etc."/"stuff goes here in actual implementation"

src/quatrex/electrostatics/meshing.py
```diff
@@ -0,0 +1,965 @@
...
+            bounding_box_0 = gmsh.model.occ.get_bounding_box(2, surface_pair[0][0])
+            bounding_box_1 = gmsh.model.occ.get_bounding_box(2, surface_pair[1][0])
+            if bounding_box_0[i] > bounding_box_1[i]:
+                surface_pair = [surface_pair[0], surface_pair[1]]
+            else:
+                surface_pair = [surface_pair[1], surface_pair[0]]
+
+            translation = np.eye(4)
+            translation[:3, -1] = lattice_vector
+            gmsh.model.mesh.set_periodic(2, *surface_pair, translation.flatten())
```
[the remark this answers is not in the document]
vetschn: If the periodic boundary surfaces are fragmented against any added geometry, there is something seriosly wrong. There can not be Dirichlet conditions on periodic surfaces, that would not make sense. @Copilot, maybe this should be enforced instead?

src/quatrex/cli/main.py
```diff
@@ -156,6 +156,51 @@ def run(
...
+    setup_context(config)
+
+    secho_header()
+
+    from quatrex.electrostatics.meshing import DeviceMesh
+
+    # Trigger mesh generation and visualization.
+    device_mesh = DeviceMesh(config)
+    device_mesh.generate_mesh()
+    device_mesh.visualize(off_screen=off_screen)
```
almaeder, line 295: 🥇

src/quatrex/core/config.py
```diff
@@ -39,10 +40,72 @@ class SCSPConfig(BaseModel):
...
 
-    mixing_factor: PositiveFloat = Field(default=0.1, le=1.0)
+    convergence_tol: PositiveFloat = 1e-3
+    r"""The convergence tolerance for the potential in the
+    Schrödinger-Poisson loop.
+
+    This is defined as the infinity norm of the difference between the
+    potential in the current iteration and the previous iteration.
+
+    $$\lVert V_{n} - V_{n-1} \rVert_{\infty} < \verb|convergence_tol|$$
```
almaeder: Why do we use the infinity norm, and should we only have an absolute tolerance?

src/quatrex/core/config.py
```diff
@@ -39,10 +40,72 @@ class SCSPConfig(BaseModel):
...
+      potential is a weighted average of the previous potential and the
+      newly computed potential. The weight is given by the
+      `mixing_factor` parameter.
+    - `"diis"`: Direct inversion in the iterative subspace (DIIS) method
+      which constructs the new potential as a linear combination of the
+      previous potentials and the newly computed potential. The
+      coefficients of the linear combination are determined by
+      minimizing the residuals of the previous potentials.
+
+    """
```
almaeder, line 75: 🙏

src/quatrex/core/config.py
```diff
@@ -89,21 +152,84 @@ class SCBAConfig(BaseModel):
...
+
+    Only used if `solving_scheme` is set to "root-finding".
+
+    """
+
+    convergence_tol: PositiveFloat = 1e-3
+    """The convergence tolerance for the root-finding scheme.
+
+    This is defined as the infinity norm of the potential update in the
+    root-finding scheme.
```
almaeder, line 223: Should we only have an abs tol with the infinite norm?

src/quatrex/core/config.py
```diff
@@ -89,21 +152,84 @@ class SCBAConfig(BaseModel):
...
+
+    Only used if `solving_scheme` is set to "root-finding".
+
+    """
+
+    density_model: Literal["single-band", "omen"] = "single-band"
```
almaeder, line 229: Maybe `legacy`? We should have some detailed explanation in the documentation about how they differ.

src/quatrex/core/config.py
```diff
@@ -89,21 +152,84 @@ class SCBAConfig(BaseModel):
...
+    """
+
+    density_model_dim: Literal[1, 2, 3] = 2
+    """The dimensionality of the system to use for the single-band
+    density model.
+
+    Only used if `solving_scheme` is set to "root-finding" and
+    `density_model` is set to "single-band".
+
+    """
```
almaeder, line 255: Should we automatically determine this, given the periodicity of the input data?

src/quatrex/core/config.py
```diff
@@ -89,21 +152,84 @@ class SCBAConfig(BaseModel):
...
+
+    initial_guess: Literal["zero", "constraints", "file"] = "file"
+    """The strategy to generate the initial guess for the potential.
+
+    - `"zero"`: Uses a zero potential as the initial guess.
+    - `"constraints"`: Solves a linear Poisson equation with the potential
+        constraints to generate the initial guess.
+    - `"file"`: Loads the initial guess from a file. The file should be
+        located in the `input_dir` and named `potential.npy`.
```
almaeder, line 266: Maybe we could do a validator to check that the file exists?

almaeder: Also, I am not sure if the default should be `file`.

src/quatrex/core/qtbm.py
```diff
@@ -552,6 +553,107 @@ def _write_outputs(self):
...
+            of states.
+        excess_hole_density : NDArray
+            The excess hole density computed from the local density of
+            states.
+        """
+
+        # Compute the spectral electron and hole densities.
+        electron_density = np.zeros((self.num_orbitals, self.electron_energies.size))
+        hole_density = np.zeros((self.num_orbitals, self.electron_energies.size))
+        for contact, ldos in self.observables.electron_ldos.items():
```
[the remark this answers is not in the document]
almaeder: Indeed, is it intended that this is on the CPU?

src/quatrex/core/qtbm.py
```diff
@@ -552,6 +553,107 @@ def _write_outputs(self):
...
+
+        # HACK: Because the potential is baked into the Hamiltonian, we
+        # need to update the Hamiltonian matrices.
+        if self.device.potential is None:
+            self.device.potential = potential
+        else:
+            delta_potential = potential - self.device.potential
+            self.device.potential = delta_potential
+
+        self.device.apply_potential()
```
almaeder: makes sense to only update with the difference

src/quatrex/core/qtbm.py
```diff
@@ -552,6 +553,107 @@ def _write_outputs(self):
...
+        for contact in self.device.contacts:
+            contact.unit_cell_hamiltonian = {}
+            contact.unit_cell_overlap = {}
+
+            contact._init_hamiltonian_overlap_matrices()
```
almaeder, line 1199: This will be removed when the OBC interface is changed in QTBM.

src/quatrex/core/scba.py
```diff
@@ -628,6 +631,87 @@ def _write_iteration_outputs(self, iteration: int):
...
+
+        Returns
+        -------
+        charge_density : NDArray
+            The computed charge density for the device.
+
+        """
+
+        self.run()
+
```
almaeder: It may not be so clear that `compute_charge_density` also runs the partial simulation. Maye call in `SCSP` separately run?

src/quatrex/core/scsp.py
```diff
@@ -0,0 +1,147 @@
...
+from quatrex.core.config import QuatrexConfig, SCSPConfig
+from quatrex.core.qtbm import QTBM
+from quatrex.core.scba import SCBA
+from quatrex.electrostatics.electrostatics import ElectrostaticSolver
+from quatrex.electrostatics.mixer import DIIS, Mixer, UnderRelaxation
+
+
+class SCSP:
+    """Self-consistent Schrödinger-Poisson solver.
+
```
almaeder, line 17: This is not yet integrated into the CLI, right?

almaeder: Structure makes a lot of sense. For the CI, we probably just need to dispatch with another `if`....

tests/quatrex/electrostatics/test_fermi_integrals.py
```diff
@@ -0,0 +1,51 @@
...
+    pytest.param(2, id="2"),
+    pytest.param(10, id="10"),
+]
+
+ORDERS = [
+    pytest.param(-1, id="-1"),
+    pytest.param(-1 / 2, id="-1/2"),
+    pytest.param(0, id="0"),
+    pytest.param(1 / 2, id="1/2"),
+]
```
almaeder, line 21: maybe into a `conftest`

src/quatrex/electrostatics/_params.py
```diff
@@ -0,0 +1,365 @@
...
+These parameters are used in the implementation of the inverse Fermi
+integrals of order -1/2 and 1/2, which are based on the minimax rational
+approximation schemes described in the following papers:
+
+- T. Fukushima, "Analytical computation of inverse Fermi-Dirac integral
+of order -1/2 by piecewise rational function approximation", 2020.
+- T. Fukushima, "Precise and fast computation of inverse Fermi-Dirac
+integral of order 1/2 by minimax rational function approximation", 2015.
+
+"""
```
almaeder, line 13: ❤️

src/quatrex/electrostatics/assembly.py
```diff
@@ -0,0 +1,270 @@
...
+        elif epsilon_r.ndim < 3:
+            if epsilon_r.ndim == 2 and epsilon_r.shape[0] != 3:
+                raise ValueError(
+                    f"Expected epsilon_r to have shape (3, ...), "
+                    f"but got {epsilon_r.shape}."
+                )
+
+            # NOTE: Broadcasting works like this for both 1D and 2D
+            # cases, since the last dimension is the number of DOFs.
+            epsilon_r = epsilon_r * np.stack(basis.N * [np.eye(3)], axis=-1)
```
almaeder, line 56: This is for anisotropy, but where are there only diagonal elements?

src/quatrex/electrostatics/solver.py
```diff
@@ -0,0 +1,301 @@
...
+        """Solves the Poisson equation for a given charge density."""
+        charge_density = self._enforce_potential_constraints(
+            -(self.fixed_density - charge_density) / epsilon_0
+        )
+        potential, __ = sparse.linalg.bicgstab(
+            self.stiffness_matrix,
+            charge_density,
+            M=self.preconditioner,
+        )
+        return self.mfc_transform.T @ potential
```
almaeder: Can be done in a later PR to enable the different solvers.

almaeder: Also, for a later RP, have the Jacobi preconditioner as a `LinearOperator` to scale instead of allocating a diagonal matrix.

almaeder (commented):
Very nice work!
This brings quatrex much closer to being a great TCAD tool.

src/quatrex/core/config.py
vetschn: This is the "legacy" way of detecting convergence. It would likely be a good idea to have a relative convergence criterion in addition to this. Would it be okay to try to address this in a follow-up PR though?

src/quatrex/core/config.py
vetschn: Again, this is the "legacy" method. I think it might be best to leave this as it is for now until we have some more real-world examples using this to gauge what a "good"/sufficient relative tolerance would be.

src/quatrex/core/config.py
vetschn: Should not be automatically determined in my opinion. 2D density models accelerate convergence in 1D structures from what i've seen, while a 1D model does not do a great job in the same system.

src/quatrex/core/config.py
vetschn: Definitely, have not had time yet to implement a better way of doing initial guess, and i used the `"file"` option to test against omen.

src/quatrex/core/qtbm.py
vetschn: I guess commented because this was flagged by the AI at some point, right? This is still a hack to subtract the original potential and set the new potential at the same time, because `apply_potential()` performs a `+=` operation to apply the potential on the Hamiltonian matrix. Pretty confusing but will have to change in the future anyway.

tests/quatrex/electrostatics/test_fermi_integrals.py
vetschn: Might do when i add more tests to the electrostatics :+1: I think these are very specific to the Fermi integral tests and are unlikely to be shared with other tests, so `conftest.py` is not really needed.

src/quatrex/electrostatics/solver.py
vetschn: Definitely! I get that using the reified Jacobi matrix preconditioner here might not be exactly as efficient as using something like
```python
diagonal = A.diagonal()
matvec = lambda x: x / diagonal
preconditioner = LinearOperator(A.shape, matvec=matvec)
```
Is this what you mean? The complexity of the operation should be the same, but it's just a little bit more efficient compared to calling `sparse.csr_matrix.dot()` or whatever, right?

src/quatrex/electrostatics/assembly.py
vetschn: This is for axis-aligned anisotropy, the case before (`if epsilon_r.ndim == 3:`) handles more general dielectric tensors including off-diagonal coupling elements.

src/quatrex/core/qtbm.py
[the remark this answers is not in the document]
vetschn: Will have to be fixed when i'm confident that all the rest works correctly.

src/quatrex/core/scba.py
vetschn: Sure, that would be a bit more explicit :+1:

src/quatrex/core/scsp.py
vetschn: Indeed, i thought i'd do `if config.scsp is not None: _run_scsp(config)`

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

vetschn:
cscs-ci run

src/quatrex/core/config.py
almaeder: We can do this together with #270.

vetschn:
Sorry about another almost unreviewable addition to the code. I'll go through the changed files one by one and leave some explanations in this comment here.


**`.cscs/Dockerfile.base`**
- @almaeder helped set up the alps ci and fixed any GPU bugs. This means including gmsh in the base Dockerfile. There are no gmsh wheels for aarch64 provided via PyPI, so this has to be built from source.
- After checking some options, looking forward it seems like [pixi](https://pixi.prefix.dev/latest/) would be a pretty good way in general to manage all dependencies.

**`.github/workflows/tests.yaml`**
- The `-x` "Exit instantly on first error or failed test." option for pytest was apparently causing hanging CI runs.

**`environment-dev.yaml`**
- The meshing and FEM stuff needs some new dependencies. I once again set up a new environment from scratch.


**`examples/**/quatrex_config.toml`**
- The electrostatics required several changes to the config structure, which i ended up refactoring a bunch, especially for the NEGF flow through the code.
- Instead of left/right fermi level, left/right temperature etc. there are now two instances of `ContactConfig` attached to the `ElectronConfig` that are used for NEGF runs. This will hopefully be brought together soonish with #114.
- Left/right band edges are no longer there. Since we compute the actual band structure, all we need are guesses for an energy sitting in the band gap (`mid_gap_energy`)
- Since the way the distance between fermi level and conduction band edges was determined for GW calculations was kind of broken before, the parameters that you can see there seem kind of strange, but these exactly reproduce the original behaviour.

**`examples/**/electron_density_*.npy`**
- Several electron and hole densities and LDOS quantities were missing a factor of 2 to account for spin degenaracy. This is important to get the correct charges.

**`examples/w90/carbon-nanotube/inputs/structure.xyz`**
- I centered the CNT structure to make it easier to define a gate etc. around it.

**`examples/w90/carbon-nanotube/qtbm-poisson/*`**
- Added a new reference test example with an electrostatics setup. The electrostatics are switched on and off by the presence or the absence of an `[scsp]` (Self-Consistent Schrödinger Poisson) section in the config.

**`mkdocs.yml`**
- Corrected the file name for the parameters here

**`pyproject.toml`**
- Added the necessary dependencies:
    - ase for reliably loading xyz files.
    - meshio to handle parsing and representation of gmsh files
    - gmsh to handle all the solid geometry and FEM meshing
    - scikit-fem for matrix assembly and constraint handling
    - pyvista to visualize meshes and regions

**`src/quatrex/bandstructure/contact.py`**
- Fermi level detection was completely refactored. It does not have to be done via the DOS of the contact, but can be directly done on the contact band structure (This is very similar to what OMEN does).
- Doping density is detected by giving in a bunch of coordinates and checking if they are all in a doped geometry region.
- Cell volume is an important factor in determining accurate charge densities. To allow a comparison with OMEN where this cell size is also handled in a bit of a weird way, _the `cell_volume` is always determined via the contact's `lattice_vector` and **not** the shape of the doping region_ (this happens in in the caller)

**`src/quatrex/cli/main.py`**
- Added a mesh command that has to be called before a run including electrostatics
- Added a scsp runner

**`src/quatrex/core/config.py`**
- Expose a bunch of options for the self-consistent Schrödinger-Poisson loop, for the electrostatics solver, the geometry and the contacts. Should hopefully be mostly clear through the docstrings.

**`src/quatrex/core/constants.py`**
- Removed the `e**2` prefactor from `epsilon_r`/`mu_r`. `epsilon_r` was unused before, but this normalization makes it better to use in the Poisson solver.


**`src/quatrex/core/qtbm.py`**
- Added charge density getter and potential setter methods to implement the new `TransportSolver` interface
- Also several smaller changes due to the new contact semantics

**`src/quatrex/core/scba.py`**
- Same as for QTBM

**`src/quatrex/core/scsp.py`**
- This handles the loop between transport and electrostatics and the mixing of the electrostatics potential between iterations
- One aspect that has to be resolved asap is the repeated tear-down/setup of QTBM. Since QTBM will probably change a bit with #284 , i didn't bother

**`src/quatrex/core/transport.py`**
- Transport solver interface. This just needs to set potentials and get excess charge densities.

**`src/quatrex/device/contact.py`**
- Added automated Fermi level detection at runtime
- Again, this class will change a bit in the future so there are some hopefully shortlived hacks still in there

**`src/quatrex/device/device.py`**
- Instantiate contact via contact config insteda of many smaller parameters

**`src/quatrex/electron/solver.py`**
- Added automated Fermi level detection at runtime
- There is a bit of strange logic about communication in there as well, that could be adressed together with #302.


**`src/quatrex/electrostatics/_params.py`**
- Cooked file with a bunch of parameters for the inverse Fermi integral approximations (see `fermi_integrals.py`)


**`src/quatrex/electrostatics/assembly.py`**
- Handles everything to do with matrix assembly on the tetrahedral mesh
- Here the multi-freedom constraints are also handled that define the periodic boundary conditions in the mesh

**`src/quatrex/electrostatics/density_response.py`**
- Density response models that model how the excess charge density  and its derivative might respond to a variations in electrostatic potential
- There is an OMEN reference implementation and a generalized version of this that takes advantage of the Fermi integrals of different orders and their inverses

**`src/quatrex/electrostatics/electrostatics.py`**
- This handles setting up a Poisson solver, getting an initial potential guess and configuring all constraints.
- It also handles the projection of the orbital-based charge density into the mesh. Currently only by assuming the orbitals are delta functions, but in the future the intent is that this class handles a more realistic real-space projection.

**`src/quatrex/electrostatics/fermi_integrals.py`**
- Evaluation of Fermi integrals and inverses for the density response models
- After some testing it seems like this can really make a difference in the speed of convergence

**`src/quatrex/electrostatics/geometry_config.py`**
- This handles the definition and resolution of geometry regions and the physical properties of those regions.

**`src/quatrex/electrostatics/meshing.py`**
- The connection between the geometry configuration and it's embedding in an actual mesh
- Based on what is present in the config and in the structure.xyz file, this builds up a mesh for the device geometry. This also handles size fields and periodic meshes.

**`src/quatrex/electrostatics/mixer.py`**
- Mixing approaches for the electrostatic potential
- Includes (adaptive) under-relaxation and a kind of "periodic" Pulay mixing that is also used in DFT sometimes.

**`src/quatrex/electrostatics/solver.py`**
- The actual Poisson solver. This is what is used in the `ElectrostaticsSolver` to actually treat the equation it has set up.
- The direct poisson solver is there more as a reference, this really does not work so well because the whole dependence between potential and charge density is super non-linear
- The non-linear solver is what OMEN also uses, it is a kind of root-finding / predictor-corrector algorithm.

**`tests/quatrex/bandstructure/test_band_edges.py`**
- Had to change this because `config.electron.conduction_band_edge` no longer exists

**`tests/quatrex/electrostatics/test_fermi_integrals.py`**
- Some basic tests for the fermi integrals.

src/quatrex/core/scba.py
```diff
@@ -685,6 +685,87 @@ def _write_iteration_outputs(self, iteration: int):
...
+        Parameters
+        ----------
+        potential : NDArray
+            The new potential values to be set in the system matrix.
+
+        """
+        if potential.shape[0] != np.sum(self.data.orbitals_per_atom):
+            potential = np.repeat(potential, self.data.orbitals_per_atom)
+
```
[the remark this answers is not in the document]
almaeder: I guess this was not changed when I tried on alps since the test didnt cover it.

almaeder (approved):
Thank you for the hard work.
After the tests pass, we can merge this and follow up with any needed fixes.

src/quatrex/bandstructure/contact.py
```diff
@@ -4,122 +4,195 @@
...
-        h_01 * xp.exp(-1j * k)[:, xp.newaxis, xp.newaxis]
-        + h_00
-        + h_10 * xp.exp(1j * k)[:, xp.newaxis, xp.newaxis]
-    )
-    if get_array_module_name(h_k) == "cupy":
-        e_k = get_device(np.linalg.eigvals(get_host(h_k)))
-    else:
-        e_k = np.linalg.eigvals(h_k)
-    return xp.sort(e_k.real, axis=1)
+    e_k = linalg.eigvalsh(h_k, s_k, compute_module="numpy")
```
maurodossena, line 54: Can the eigenvalue problem be solved also on the GPU?

src/quatrex/bandstructure/contact.py
```diff
@@ -4,122 +4,195 @@
...
+
+def contact_doping_density(
+    coordinates: NDArray, geometry_regions: list[Region]
+) -> float:
+    """Computes the doping density of a device contact.
+
+    This function checks which geometry region the contact coordinates
+    are located in and returns the corresponding doping density. If the
+    contact is located in multiple regions, the doping density of the
+    first detected region is returned. If the contact is not located in
```
maurodossena, line 98: Should we support the contact being located in multiple doping regions? We could print a warning or raise an exception in that case.

src/quatrex/core/config.py
maurodossena: I would also replace `OMEN` with `legacy`

src/quatrex/core/config.py
```diff
@@ -89,21 +164,106 @@ class SCBAConfig(BaseModel):
...
+    Only used if `solving_scheme` is set to "root-finding".
+
+    """
+
+    density_model_dim: Literal[1, 2, 3] = 2
+    """The dimensionality of the system to use for the single-band
+    density model.
+
+    Only used if `solving_scheme` is set to "root-finding" and
+    `density_model` is set to "single-band".
```
maurodossena, line 253: Maybe add couple of lines explaining that 2 is expected to give the best results and that can be also used  for 1D and 3D geometries.

src/quatrex/core/config.py
```diff
@@ -89,21 +164,106 @@ class SCBAConfig(BaseModel):
...
+    """The default relative permittivity to use for the Poisson solver.
+
+    This is used as a fallback for regions that do not have a specified
+    relative permittivity.
 
-    num_orbitals_per_atom: dict[str, int] = Field(default_factory=dict)
+    """
+
+    electron_affinity: float | None = None
+    """The electron affinity of the device.
```
maurodossena: Better to specify `electron affinity of the semiconductor channel`

src/quatrex/core/qtbm.py
```diff
@@ -552,6 +558,111 @@ def _write_outputs(self):
...
+        self.device.apply_potential()
+        for contact in self.device.contacts:
+            contact._init_hamiltonian_overlap_matrices()
+
+    def get_charge_density(self) -> NDArray:
+        """Gets the charge density from the QTBM calculation.
+
+        This method runs the full QTBM calculation and then integrates
+        the local density of states to obtain the charge density. This
+        is typically used in self-consistent calculations where the
```
maurodossena: The QTBM calculation runs in another method right? This method seems to just integrate the local DOS to obtain the charge density.

src/quatrex/core/scba.py
```diff
@@ -633,6 +641,91 @@ def _write_iteration_outputs(self, iteration: int):
...
+        self.electron_solver.potential = potential
+
+    def get_charge_density(self) -> NDArray:
+        """Gets the charge density.
+
+        This runs the SCBA to convergence (or to the maximum number of
+        iterations) and then computes the charge density from the
```
maurodossena: Same comment I had for QTBM. This method just computes the charge density right?

src/quatrex/device/contact.py
```diff
@@ -263,6 +252,74 @@ def __init__(
...
+                self.mid_gap_energy,
+                self.delta_fermi_level_conduction_band,
+            ) = self._compute_fermi_level(
+                num_kpoints_transport=contact_config.num_kpoints_transport,
+                kpoints_transverse=kpoints_transverse,
+                mid_gap_energy=contact_config.mid_gap_energy,
+                temperature=contact_config.temperature,
+                doping_density=doping_density,
+                cell_volume=np.abs(np.linalg.det(self.lattice_vectors)),
+            )
```
maurodossena, line 379: Will it work for transverse periodic contacts? I guess we need to fold the kpoints_transverse grid into a finer one in case.

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

maurodossena (approved):

src/quatrex/bandstructure/contact.py
vetschn: I think the best way to handle overlapping regions like this would be some sort of [z-ordering](https://en.wikipedia.org/wiki/Z-order) tags or something. We can discuss if this is worth it :+1: 

See #318

src/quatrex/core/qtbm.py
vetschn: Good catch! Used to be the case but i forgot to change the docstring :+1:

src/quatrex/device/contact.py
vetschn: Looking at it again after a while now but it looks like i'm using `_construct_contact_matrix` to get the full matrices from the unit cell ones. I guess it will kinda work, but the contact eigenvalue problem can actually be quite a lot more expensive than it needs to be :grimacing: Also not sure if the doping will get normalized correctly with the cell size.

See #317

src/quatrex/bandstructure/contact.py
vetschn: It can absolutely. Just did this for testing. See #321

vetschn:
cscs-ci run

src/quatrex/bandstructure/contact.py
```diff
@@ -4,122 +4,195 @@
...
+from quatrex.electrostatics.meshing import inside_shape
+
+
+def _real_to_kspace(
+    kpoints: NDArray, a_xx: tuple[NDArray, NDArray, NDArray] | None
+) -> NDArray:
+    """Transforms real-space operator blocks to k-space."""
+    a_10, a_00, a_01 = a_xx
+    a_k = (
+        np.einsum("k, ...-> k...", np.exp(-1j * kpoints), a_10)
```
alexnick83, line 19: Shouldn't this be xp? I guess if eigvalsh (below) has to be done on the CPU (AMD), you will still have to copy the result back to the host, but I guess this is fine? How big are the `a_xx` matrices?

src/quatrex/bandstructure/contact.py
```diff
@@ -4,122 +4,195 @@
...
+
+    Returns
+    -------
+    doping_density : float
+        The doping density in Å^-3.
+
+    """
+    for region in geometry_regions:
+        if not isinstance(region.properties, VolumeProperties):
+            continue
```
alexnick83, line 116: Are there legitimate cases where `region.properties` is not a `VolumeProperties` object or is this essentially a None-check?

src/quatrex/bandstructure/contact.py
```diff
@@ -4,122 +4,195 @@
...
+
+        if (
+            region.properties.donor_concentration == 0.0
+            and region.properties.acceptor_concentration == 0.0
+        ):
+            continue
+
+        if np.all(inside_shape(coordinates, region.shape)):
```
alexnick83, line 124: The docstring says that if the contact spans multiple regions, then it will return the doping density of the first detected one. This may be a matter of interpretation, but this sounds to me as if such a contact will never satisfy this check, since not all its coordinates will be inside the region. Please note, I haven't seen yet what `inside_shape` does, but I am assuming that it returns a boolean container of the same class and size as `coordinates` with value True for individual coordinates that are inside `region.shape` and False otherwise.

alexnick83: If the check is never satisfied, then the docstring is wrong.

src/quatrex/bandstructure/contact.py
```diff
@@ -4,122 +4,195 @@
...
-        The Fermi level.
+    fermi_level : float
+        The Fermi level in eV.
 
     """
-    dE = energies[1] - energies[0]
+
+    # NOTE: int() does a implicit copy to the host
+    # which is need since the input for split needs to be a int
+    num_valence_bands = int((e_k < mid_gap_energy).sum(axis=1).max())
```
alexnick83: It may be helpful to describe somewhere the definition based on the input. `e_k` is the eigenvalues by k-points and bands (energy points?). So you find which eigenvalues are smaller than the mid-gap energy, you  count how many such eigenvalues exist per k-point, and then you return the max number per k-point?

src/quatrex/core/config.py
```diff
@@ -105,21 +180,111 @@ class SCBAConfig(BaseModel):
...
+
+    density_model: Literal["single-band", "omen"] = "single-band"
+    """The density model to use for the root-finding scheme.
+
+    - `"single-band"`: Uses a simple single-band density model where the
+      charge density is computed from the potential using a single-band
+      approximation.
+    - `"omen"`: Uses the density model from the OMEN code. This is
+      almost identical to the single-band model with `density_model_dim
+      = 2`. However, it uses slightly different physical constants, i.e.
```
alexnick83: `i.e., ` (missing comma).
What does it mean exactly that OMEN uses slightly different physical constants? Is it a numerical issue, or does it use a different formulation? Is this just for validation against OMEN?

src/quatrex/core/config.py
```diff
@@ -105,21 +180,111 @@ class SCBAConfig(BaseModel):
...
+    `density_model` is set to "single-band".
 
-    num_orbitals_per_atom: dict[str, int] = Field(default_factory=dict)
+    """
+
+    initial_guess: Literal["zero", "constraints", "file"] = "zero"
+    """The strategy to generate the initial guess for the potential.
+
+    - `"zero"`: Uses a zero potential as the initial guess.
+    - `"constraints"`: Solves a linear Poisson equation with the
```
alexnick83, line 261: When exactly would we use "constraints"? It looks to me that we start either with zero or file. If we add support for restarting, then we save the last potential to file and use that for the restart.

src/quatrex/core/config.py
```diff
@@ -488,6 +653,122 @@ class LyapunovConfig(BaseModel):
...
+        `"wf"` formalism.
+
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    name: str
+    """A unique name for the contact."""
+
+    origin: tuple[float, float, float] = (0.0, 0.0, 0.0)
```
alexnick83, line 671: I guess unrelated to the PR itself, but do we explain somewhere where exactly (0, 0, 0) is located? Is it at the edge of the simulation domain, i.e., you cannot have a negative coordinate? Does it depend on the device's geometry?

src/quatrex/core/config.py
```diff
@@ -488,6 +653,122 @@ class LyapunovConfig(BaseModel):
...
+    In `"wf"` simulations this is used to automatically determine the
+    orbitals that belong to this contact.
+
+    The volume of the contact cell is also used to determine the Fermi
+    level of the contact from its doping density and the density of
+    states of its band structure.
+
+    """
+
+    direction: Literal["a", "b", "c"] | None = None
```
alexnick83, line 702: What does `None` imply here? That the contact is not periodic or that it consists of a single cell?

src/quatrex/core/config.py
```diff
@@ -488,6 +653,122 @@ class LyapunovConfig(BaseModel):
...
+    !!! warning
+
+        This parameter is currently only used in the `"wf"` formalism.
+
+    """
+
+    fermi_level: float | None = None
+    """The Fermi level of the contact.
+
+    If not set, the Fermi level is automatically determined from the
```
alexnick83, line 717: Based on the descriptions, is this for backward compatibility?

src/quatrex/core/config.py
```diff
@@ -488,6 +653,122 @@ class LyapunovConfig(BaseModel):
...
+
+    mid_gap_energy: float | None = None
+    """An energy lying somewhere in the band gap of the contact.
+
+    This is used to separate conduction from valence band states, which
+    is necessary to automatically determine a contact's Fermi level, and
+    to compute the excess carrier density that is used in computing the
+    electrostatic potential in the Poisson solver.
+
+    This is also necessary when band-edge tracking is enabled in
```
alexnick83, line 734: I suppose the mid-gap energy in the transport model is separate and concerns the transistor channel, which may have different band edges than the contacts?

src/quatrex/core/config.py
```diff
@@ -488,6 +653,122 @@ class LyapunovConfig(BaseModel):
...
+
+    """
+
+    num_kpoints_transport: int = 51
+    """Number of k-points to use for contact band structure calculation.
+
+    This is used when automatically determining the Fermi level of the
+    contact from its band structure. The k-point grid along the
+    transverse directions are determined from the
+    [`kpoint_grid`][kpoint_grid] parameter.
```
alexnick83: Second kpoint_grid lacks ``. Are these grids the same or different?

src/quatrex/core/constants.py
```diff
@@ -2,11 +2,11 @@
...
 
-c_0 = 1e10 * physical_constants["speed of light in vacuum"][0]  # angstrom / s
+c_0 = 1e10 * physical_constants["speed of light in vacuum"][0]  # Å / s
 e = physical_constants["elementary charge"][0]  # C
 hbar = physical_constants["reduced Planck constant in eV s"][0]  # eV s
 h = physical_constants["Planck constant in eV s"][0]  # eV s
 alpha = physical_constants["fine-structure constant"][0]  # dimensionless
-epsilon_0 = e**2 / (2 * alpha * h * c_0)
-mu_0 = 2 * alpha * h / (c_0 * e**2)
+epsilon_0 = 1 / (2 * alpha * h * c_0)  # 1 / (eV Å)
```
alexnick83, line 10: I assume the change implies that we always use these constants as `epsion_0 * mu_0`, or are the missing `e**2` factors compensated for in the code?

src/quatrex/core/scba.py
```diff
@@ -685,6 +685,87 @@ def _write_iteration_outputs(self, iteration: int):
...
+        hole_density = self.observables.hole_density.copy()
+
+        mask = self.electron_energies[:, None] > mid_gap_energy
+        electron_density[~mask] = 0
+        hole_density[mask] = 0
+
+        excess_electron_density = np.trapezoid(
+            electron_density, self.electron_energies, axis=0
+        )
+        excess_hole_density = np.trapezoid(hole_density, self.electron_energies, axis=0)
```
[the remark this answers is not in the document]
alexnick83: I think this was also the case in QTBM but was fixed there.

src/quatrex/core/scba.py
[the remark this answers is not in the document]
alexnick83: See above

src/quatrex/core/scba.py
```diff
@@ -637,6 +645,90 @@ def _write_iteration_outputs(self, iteration: int):
...
+        -------
+        charge_density : NDArray
+            The computed charge density for the device.
+
+        """
+        electron_density, hole_density = self._compute_excess_charge_densities()
+        charge_density = electron_density - hole_density
+
+        # From orbital to atom resolved charge density.
+        orbital_offsets = np.hstack(([0], np.cumsum(self.data.orbitals_per_atom)))
```
alexnick83, line 727: Do we want these on the host?

src/quatrex/core/transport.py
```diff
@@ -0,0 +1,24 @@
...
+from qttools import NDArray
+
+
+class TransportSolver(ABC):
+    """Abstract base class for transport solvers."""
+
+    @abstractmethod
+    def set_potential(self, potential: NDArray) -> None:
+        """Sets the potential for the transport solver."""
+        ...
```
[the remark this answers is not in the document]
alexnick83: Using an ellipsis is absolutely fine. It only makes a difference if you do something like `super().set_potential(...)`. It may also be an issue with generating documentation, but I haven't checked. @vetschn do you foresee any of those two potential issues? Otherwise, I would dismiss the finding

src/quatrex/core/transport.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @abstractmethod
+    def get_charge_density(self) -> NDArray:
+        """Gets the charge density from the transport solver."""
+        ...
```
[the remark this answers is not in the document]
alexnick83: See above.

src/quatrex/core/transport.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @abstractmethod
+    def run(self) -> None:
+        """Solves the transport problem."""
+        ...
```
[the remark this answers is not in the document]
alexnick83: See above.

src/quatrex/device/contact.py
```diff
@@ -338,6 +327,75 @@ def __init__(
...
+
+            doping_density = contact_doping_density(
+                coordinates=device.atom_coordinates[self.origin_atom_indices],
+                geometry_regions=device.config.device.geometry.regions,
+            )
+
+            if comm.rank == 0:
+                print(f"    Doping density: {doping_density} Å^-3", flush=True)
+
+            # HACK: If there is a potential present in the input folder,
```
alexnick83, line 353: Can this be solved by changing how/when the potential is added to the Hamiltonian? I guess for future PR ...

src/quatrex/device/contact.py
```diff
@@ -1003,6 +1061,100 @@ def _configure_nevp(self, obc_config: OBCConfig, nevp_config: NEVPConfig) -> NEV
...
+        fermi_level : float
+            The computed Fermi level in eV.
+        mid_gap_energy : float
+            The recomputed mid-gap energy based on the band structure.
+        delta_fermi_level_conduction_band : float
+            The energy difference between the Fermi level and the
+            conduction band edge.
+
+        """
+        # NOTE: I'm not 100% sure if i should be including the endpoint
```
alexnick83: In NumPy 2.4, endpoint=True is the default, so you are indeed including the same momentum twice. Unless you are compensating for that later (half weight for pi and -pi), this is wrong, no? However, why include pi twice in the first place?

src/quatrex/electron/solver.py
```diff
@@ -133,32 +140,83 @@ def __init__(
...
-        self.left_mid_gap_energy = 0.5 * (
-            config.electron.conduction_band_edge + config.electron.valence_band_edge
-        )
-        self.left_fermi_level = config.electron.left_fermi_level
-        self.right_fermi_level = config.electron.right_fermi_level
 
-        potential = self.left_fermi_level - self.right_fermi_level
-        self.right_mid_gap_energy = self.left_mid_gap_energy - potential
-        self.temperature = config.electron.temperature
+        orbitals_per_atom = [
```
alexnick83, line 145: Is this the same as in the init of SCBAData?

src/quatrex/electrostatics/__init__.py
```diff
@@ -0,0 +1 @@
+# Copyright (c) 2024 ETH Zurich and the authors of the quatrex package.
```
alexnick83:
```suggestion
# Copyright (c) 2024-2026 ETH Zurich and the authors of the quatrex package.
```

src/quatrex/core/scsp.py
```diff
@@ -0,0 +1,156 @@
+import os
```
alexnick83, line 3:
```suggestion
# Copyright (c) 2024-2026 ETH Zurich and the authors of the quatrex package.

import os
```

src/quatrex/electrostatics/assembly.py
```diff
@@ -0,0 +1,270 @@
...
+        corresponding boundary is periodic.
+
+    Returns
+    -------
+    sparse.csr_matrix
+        The transformation matrix that maps the degrees of freedom of
+        the mesh.
+
+    """
+    shape = (basis.N, basis.N)
```
alexnick83, line 139: How big is N? Do we always want this computation on the host? I guess the main constraint is skfem (only CPU code?).

src/quatrex/electrostatics/density_response.py
```diff
@@ -0,0 +1,352 @@
...
+    raise ValueError("Invalid dimensionality. Must be 1, 2, or 3.")
+
+
+class DensityModel(ABC):
+    """Abstract base class for density response models."""
+
+    @abstractmethod
+    def density(self, potential: NDArray) -> NDArray:
+        """Computes the density for a given potential."""
+        ...
```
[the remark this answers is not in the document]
alexnick83: Same issue with other abstractmethods. Potentially dismiss.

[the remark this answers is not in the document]
alexnick83: Same issue with other abstractmethods. Potentially dismiss.

src/quatrex/electrostatics/density_response.py
```diff
@@ -0,0 +1,352 @@
...
+
+    @abstractmethod
+    def density_derivative(self, potential: NDArray) -> NDArray:
+        """Computes the derivative of the density with respect to the potential."""
+        ...
```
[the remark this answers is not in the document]
alexnick83: Same issue with other abstractmethods. Potentially dismiss.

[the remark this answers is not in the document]
alexnick83: Same issue with other abstractmethods. Potentially dismiss.

src/quatrex/electrostatics/fermi_integrals.py
```diff
@@ -0,0 +1,352 @@
...
+    Returns
+    -------
+    gamma_n : float
+        The value of the gamma function at n.
+
+    """
+    if n in _gamma_values:
+        return _gamma_values[n]
+
+    return sp.special.gamma(n)
```
alexnick83, line 39: Unsure how slow this is or how often it is called. However, there exists cupyx.scipy.special.gamma, and we can also cache values.

src/quatrex/electrostatics/fermi_integrals.py
```diff
@@ -0,0 +1,352 @@
...
+        # NOTE: Factor two from variable transformation, and the gamma
+        # function in the denominator of the -1/2 order Fermi
+        # integral, which is needed to compute the derivative for
+        # k=-3/2.
+        return np.trapezoid(f, t, axis=1) * 2 / _gamma((k + 1) + 1)
+
+    a, b = -4.5, 5.0
+    t = np.linspace(a, b, num_quad_points)
+    x = np.exp(t - np.exp(-t))
+    f = x * (1 + np.exp(-t)) * x**k / (1 + np.exp(x - eta[..., np.newaxis]))
```
alexnick83, line 106: Do we want all these on the host?

alexnick83 (approved):
Some questions. I see a lot of `np` instead of `xp`, but unclear if we want stuff on the host or is it WIP. Nothing blocking. I guess subsequent PRs will clarify things.

src/quatrex/bandstructure/contact.py
vetschn: `a_xx` are currently the size of the full contact blocks (this could be improved, see #319). Not huge but this would still be worth a look, perhaps in an "optimization" pass over the code 🤷

src/quatrex/bandstructure/contact.py
vetschn: Regions can be 2D surfaces (gates) with `region.properties` being `SurfaceProperties`. These only have info about voltage, no doping or anything like that

src/quatrex/bandstructure/contact.py
vetschn: It's less that the contact spans multiple regions, but rather that multiple regions overlap each other and the contact coordinates are all inside more than one region. See also @maurodossena's comment here https://github.com/quatrex/quatrex/pull/294/changes#r3341109944

src/quatrex/bandstructure/contact.py
vetschn: Sorry, not quite sure what you mean here, but i can definitely see that it's not very clear what this function is supposed to do just from the docstring.

So the eigenvalues $\varepsilon(\mathbf{k})$ are the band structure of the contact. This line then counts how many bands are consistently below the `mid_gap_energy`, these are the valence bands. `(e_k < mid_gap_energy).sum(axis=1)` should actually contain the same number of bands for all kpoints, otherwise things are off. The call to `.max()` could also be just `[0]` as well. Not sure which is more meaningful though.

src/quatrex/core/config.py
vetschn: Yep, this is/was for validation against OMEN. Not a different formulation, just that OMEN uses `k_B = 1.38e-23 / 1.6022e-19 = 8.613156909249781e-05` (eV/K) while the [codata value](https://physics.nist.gov/cuu/Constants/index.html) used in quatrex is `k_B = 8.617333262145179e-05`

src/quatrex/core/config.py
vetschn: I've seen that using the constraints to generate a more physical initial guess can actually reduce the number of Schrödinger-Poisson iterations a lot. This is also similar to what OMEN does. Maybe this should actually be the default value

src/quatrex/core/config.py
vetschn: Good point. It depends on what is given in the `structure.xyz` file

src/quatrex/core/config.py
vetschn: I added the `None` to make this optional for the case where the contact definition is used `"negf"` calculations. It does not have any other implications. Perhaps worth adding a validator to check that it is not None when `formalism = "wf"`

src/quatrex/core/config.py
vetschn: I would say more to give the user a way of overriding the automated detection, especially when not doing `scsp`. I can see that this could be useful, e.g. to inject a bunch of current and try out some things. (similar option exists in OMEN)

src/quatrex/core/config.py
vetschn: The mid-gap energy is now only in the contacts, i guess you left this comment before that was changed?

src/quatrex/core/config.py
vetschn: They are the same grid and actually have to be in `scsp`, otherwise the computed charge density and the Fermi-levels would be inconsistent

src/quatrex/core/constants.py
vetschn: The epsilon/mu were not actually used in the code itself and only in scripts i had locally. If we include the automatic assembly of the Coulomb matrix, then the `e**2` factor will have to be compensated (see #274)

src/quatrex/core/transport.py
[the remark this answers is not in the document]
vetschn: No i don't see this issue at all, i actually think `...` is more meaningful as "stuff will go here in an actual implementation", but the bot absolutely seems to hate ellipsis

src/quatrex/device/contact.py
vetschn: Yep, this is related to copies of the contact Hamiltonian/overlap matrices existing in the `Contact` class. I think the plan is to address this with something like #114 (or somewhere else @almaeder ?)

almaeder: It should be addressed when sharing the device and contact between SCBA and QTBM.

src/quatrex/device/contact.py
vetschn: I will have a look again, i think it should even work to just do the range $[0, \pi]$ since it should be periodic along that direction anyway. Since in the test devices the bands at $\pm \pi$ are already shifted far away from the band gaps at $0$, the Fermi level is not different than the ones from OMEN, so i will look at another example to validate

src/quatrex/electrostatics/assembly.py
vetschn: N is the number of nodes in the FEM mesh, this can vary a lot with the coarseness of the mesh and with what other regions besides the actual atoms are in the simulation domain. For the CNT example (768 orbital sites) there are ~30k nodes in the mesh.

`skfem` actually seemed quite efficient for all the devices i tested on. We can try to push it a bit with very fine meshes and maybe look at a mechanism to save out the assembled stiffness matrix to run I-V characteristics where the assembly should then be amortized pretty quickly

src/quatrex/electrostatics/fermi_integrals.py
vetschn: I think for our purposes this will actually always hit the little `_gamma_values` cache

src/quatrex/electrostatics/fermi_integrals.py
vetschn: Leaving a general answer: You left several good notes on whether stuff needs to be/happen on device or host. Would it be okay to open a subsequent issue and PR to investigate and optimize GPU-related things in the electrostatics part of the code?

src/quatrex/bandstructure/contact.py
alexnick83: e(k) is num-k-points by num-bands. If all k-points are supposed to have the same number of bands with eigenvalues below the mid-gap energy, then I see two cases:
- You trust that this never fails, so you sum up the band of only a single k-point, e.g., `return (e_k[0] < mid_gap_energy).sum()`
- You do not trust this, and you validate it. Then `nbands = (e_k < mid_gap_energy).sum(axis=1); assert np.all(nbands == nbdans[0])` or something like that.

Now, it looks like you both do "redundant" (may not actually be) computation and fail to assert that things are not off.

src/quatrex/core/config.py
alexnick83: So, quatrex itself does not enforce a specific "simulation" origin. Is everything relative to the input structure file?

src/quatrex/bandstructure/contact.py
alexnick83: So, if I understand correctly, the contacts are always within a single region (or at least this is what is actually supported now, based on the linked comment). However, other regions may overlap with this single region at points where the contact lies. Therefore, there should always exist a region that contains the whole contact. Am I correct?

vetschn: So the user can currently define doping regions at  arbitrary places in the simulation domain and these regions currently have no priorities among each other.

The set of orbital coordinates making up the contact does not have to be contained in any of these regions, in which case the doping density for this set of coordinates defaults to `0.0`.

If there are two overlapping doping regions defined, and the contact orbital coordinates are a subset of both of these regions, then only the first checked region determines the doping density, so there is a bit of undefined behavior.

~~I should probably add a check to make sure no regions overlap~~ 

(It is not trivial at all to check if two regions with arbitrary shapes overlap 😕 )

src/quatrex/core/config.py
vetschn: Yes, exactly

src/quatrex/electron/solver.py
vetschn: Yes it is 💀  It is actually also recomputed multiple times in QTBM. Not quite sure what would be a good fix for this, i guess some global class with Device info like mentioned in #117

vetschn:
cscs-ci run

src/quatrex/electrostatics/fermi_integrals.py
vetschn: See #321

PR #301, opened 2026-05-21
Unify the halo communication

almaeder:
Currently, we have two implementations with nccl and mpi.
This PR should unify them into one by having `send` and `recv` in the communicator.


almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

.cscs/alps-ci.yaml
```diff
@@ -47,25 +47,25 @@ build quatrex:
 test quatrex single:
   extends: .test-common
   script:
-    - coverage run --source=src --parallel-mode -m pytest tests/quatrex
+    - coverage run --source=src --parallel-mode -m pytest -vv tests/quatrex
```
vetschn, line 50: Is this for debugging or what was the intention here?

examples/w90/carbon-nanotube/gw-dist/reference-outputs/device_current_0.npy
vetschn, line 1: A lot of difference? Or just because full solver now?

examples/w90/carbon-nanotube/gw-dist/quatrex_config.toml
vetschn, line 1: The changes here are because of the RNG in Beyn?

src/qttools/comm/comm.py
```diff
@@ -238,8 +244,25 @@ def _check_bufs_consistent(self, sendbuf: NDArray, recvbuf: NDArray):
...
+        backend : str, optional
+            The backend to use for the communication. If None, the default
+            backend will be used.
+
+        """
+        if self._group_start_called:
+            raise RuntimeError(
+                "This cannot be called between group_start and group_end."
+            )
+
```
vetschn, line 265: Why? This is a restriction we want to impose or something from the underlying comm backends?
From the NCCL docs it sounds like grouping collectives like this is even encouraged

Is it because cupy's NCCLCommunicator implements collectives by goruping point-to-point calls?

src/qttools/comm/comm.py
vetschn, line 1: Looks very nice 👌 The non-blocking comm methods are currently not used anywhere in the code. These are just there for completeness in case we want to use them at some point?

src/qttools/datastructures/routines.py
vetschn: Complete nit but i always prefer to have an early return instead of these `if` guards

src/qttools/utils/sparse_utils.py
vetschn, line 1: It looks like `product_sparsity_pattern_dsdbsparse()` is the only method in this module that is actually used. Maybe just remove the others?

tests/quatrex/reference/test_reference.py
vetschn, line 1: A lot better 👍

vetschn (approved):
Looks pretty good to me 👍

.cscs/alps-ci.yaml
almaeder: Yes, we mostly care about the CI when the tests fail, and it is best to see more verbosely where.
This is nice for when one creates deadlocks on a regular basis (as I do).
I do not think there is a demerit in keeping it turned on, right?

examples/w90/carbon-nanotube/gw-dist/reference-outputs/device_current_0.npy
almaeder: Mostly because of full and because it was randomly failing both locally and on Alps.
The test was very sensitive to how many ranks were used to test.

examples/w90/carbon-nanotube/gw-dist/quatrex_config.toml
almaeder: Yes

src/qttools/comm/comm.py
almaeder: Initially, to impose restrictions on ourselves.
I believe the abstraction should then be to call `i`collectives with group start + end, but we would need to check `NCCLCommunicator` again.

The feature is sadly not yet used since I didn't want to do another `if/else`.

src/qttools/comm/comm.py
almaeder: The non-blocking was added before I decided that the feature for the sandwich should come later.
I believe this feature should be integrated with a performance check to see how well it works on Alps / Frontier.

src/qttools/datastructures/routines.py
```diff
@@ -629,38 +242,38 @@ def _recv_from_next():
...
     # Receive halo blocks from previous rank
     def _recv_from_previous():
         if start_block > 0:
             for i in range(start_block, min(num_blocks, start_block + a_off)):
                 for j in range(max(0, i - a_off), min(start_block, i + a_off + 1)):
                     a[i, j] = xp.empty((a_ssz) + (bsz[i], bsz[j]), dtype=dtype)
-                    nccl_comm.recv(a[i, j], rank - 1)
+                    comm.block.recv(a[i, j], rank - 1)
             for j in range(start_block, min(num_blocks, start_block + b_off)):
                 for i in range(max(0, j - b_off), min(start_block, i + b_off + 1)):
```
almaeder: @vetschn isn't this a bug? `i` is used in the bounds of the range while it is also `for i in`.

vetschn: Strange, i think it should be `j` in the range

vetschn: Indeed a bug, but this case is never actually hit in the code or in the tests. I think this only ever becomes a problem when multiplying matrices that are more than tri-diagonal (the `b` matrix is always tri-diagonal in our applications, also in the sandwich case)

almaeder: I will fix it. Should we include further tests in this PR or `trust`?

vetschn: I guess that would mean adding test cases with `a` and `b` both penta-diagonal at least, right? An alternative could be to raise an exception if `b` has too many off-diagonals and note that this is untested :skull:

almaeder:
cscs-ci run

vetschn:
cscs-ci run

Issue #275, opened 2026-04-14
Profiler deadlock


almaeder:
The current profiler can lead to a deadlock if there is an exception inside the called function.
This is due to the all-to-all after the `try`.
We should fix this by returning and giving the expected error message.

almaeder:
This should be fixed by #304.

PR #304, opened 2026-05-27
Profiler deadlock

almaeder:
This PR addresses #275.
It is fixed by removing the `finally` inside the profiler.
Further, logic is included in the CLI to abort the program in case of failure.
It was observed that the program hangs on Alps when throwing an exception.

### DIFF
--- a/src/qttools/profiling/profiler.py
+++ b/src/qttools/profiling/profiler.py
@@ -389,44 +389,45 @@ class Profiler:
         timestamp = time.time()
 
         # NOTE: We maybe need to barrier before starting the timer
+        exception_raised = False
+        try:
+            if xp.__name__ == "cupy":
+                xp.cuda.runtime.deviceSynchronize()
+                if NVTX_AVAILABLE:
+                    xp.cuda.nvtx.RangePush(label)
+            start_time = time.perf_counter()
 
-        if xp.__name__ == "cupy":
-            xp.cuda.runtime.deviceSynchronize()
-            if NVTX_AVAILABLE:
-                xp.cuda.nvtx.RangePush(label)
-        start_time = time.perf_counter()
+            yield
 
-        # NOTE: There is not finally here since this would
-        # potentially lead to deadlocks.
-        # The only issue is that if the profiled code raises an exception,
-        # the depth will not be decreased and the profiling data will be inconsistent.
-        # However, we should always crash when profiled code raises an exception,
-        # so this should not be a problem in practice.
-        yield
+        except Exception:
+            exception_raised = True
+            raise
 
-        if xp.__name__ == "cupy":
-            xp.cuda.runtime.deviceSynchronize()
-            if NVTX_AVAILABLE:
-                xp.cuda.nvtx.RangePop()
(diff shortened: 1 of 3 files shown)


almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
The Alps tests currently fail since GPU-aware MPI is not available. Due to calling into `cli_run`, this leads to the tests aborting.

almaeder:
cscs-ci run

src/quatrex/cli/main.py
```diff
@@ -102,6 +104,42 @@ def _run_negf(config):
...
+    # Force MPI to abort in the case of an exception
+    # to avoid hanging processes.
+    try:
+        full_traceback = "".join(traceback.format_exception(e))
+
+        error_msg = (
+            f"\n[RANK {comm.rank}] !!! CRITICAL EXCEPTION !!!\n" f"{full_traceback}\n"
+        )
+
+        os.write(2, error_msg.encode("utf-8"))
```
vetschn: A bit better to use `sys.stderr.write()` for this i think, i didn't know what `2` meant before googling

src/qttools/profiling/profiler.py
vetschn, line 1: Instead of removing the try/finally, could you not just add an except clause like:
```python
exception_raised = False
try:
    ...

except Exception:
    exception_raised = True
    raise

finally:
    ...
    call_time = time.perf_counter() - start_time
    if comm is not None and QTX_PROFILE_COMM_SYNC and not exception_raised:
        comm.barrier()
        after_barrier_time = time.perf_counter() - start_time
    else:
        after_barrier_time = call_time
    ...

```
Or am i misunderstanding where the hang was happening exactly?

src/quatrex/cli/main.py
```diff
@@ -114,46 +152,62 @@ def run(
...
+                    "'./quatrex_config.toml' does not exist."
+                )
+
+        # If a directory is provided, look for the config file inside.
+        if config.is_dir():
+            config = config / "quatrex_config.toml"
+            if not config.exists():
+                raise BadArgumentUsage(
+                    f"No quatrex configuration file found in directory: {config.parent}"
+                )
```
vetschn: Consider splitting the body of try into a `_resolve_config_path` method and something like `_run_quatrex`

src/quatrex/cli/main.py
```diff
@@ -114,46 +152,62 @@ def run(
...
+            )
 
-    if config.outputs.save_profiling_results:
-        profiler.dump_stats()
+        if config.outputs.save_profiling_results:
+            profiler.dump_stats()
+    except Exception as e:
+        if abort_on_exception:
+            _abort_quatrex(e)
+        else:
```
vetschn: No need for else here since `_abort_quatrex()` is `NoReturn`

almaeder:
cscs-ci run

vetschn (approved):
Great stuff 👌

vetschn:
cscs-ci run

PR #306, opened 2026-06-01
Scba batching

almaeder:
This PR should solve issue #255.
Partial batching is implemented in both the electron and the coulomb solvers.
This is done by extending the dataclass to enable views of views.
The result is a bit cooked, but we hopefully clean it up with #222.

This PR will be followed up by other PRs to include the other optimizations from SC25. I believe this should be done before refactoring the `dsdbsparse`.

PR #307, opened 2026-06-02
Update CI

### LINKED ISSUE #290: Slow test suite
Currently, the tests take up to 30min which is mainly due to the reference tests with MPI.
We should investigate the cause and try to bring down the test time.

almaeder:
- The Alps CI is now on the normal queue since there is a job limit on the debug queue.
- Remove `-x` from the tests to remove deadlocks in the MPI tests.
- Add `-vv` to the tests for more debug information.
- Reduce the number of ranks on the GitHub runner from 6 to 3 since there are only 4 cores.

### DIFF
--- a/.cscs/alps-ci.yaml
+++ b/.cscs/alps-ci.yaml
@@ -42,22 +42,22 @@ build quatrex:
     SLURM_NTASKS: 1
     SLURM_GPUS_PER_TASK: 1
     SLURM_TIMELIMIT: '00:20:00'
-    SLURM_PARTITION: debug
+    SLURM_PARTITION: normal
 
 test quatrex single:
   extends: .test-common
   script:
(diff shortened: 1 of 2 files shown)


almaeder:
cscs-ci run

almaeder:
@vetschn `tests/quatrex/reference/test_reference.py::test_distributed[w90-carbon-nanotube-gw-dist` is now failing with the github runner, but this is due to dev still using beyn with that example.
#301 will fix this.

vetschn:
~~@almaeder Would you also reduce the number of ranks in the github ci runner in this PR directly, or would this cause too many side-effects? Maybe to 4 ranks, since it has 4 cores?~~

I cannot read, sorry

vetschn (approved):
Comparing to the #306 branch, this looks to me like this shaves off like 5 minutes from the distributed tests 🎖️

PR #312, opened 2026-06-03
Added functions to load and store hdf5 files + updated examples

### LINKED ISSUE #311: .mat supports files up to 2GB
Scipy functions `savemat` and `loadmat` only support v7.2 .mat files, with their size capped to 2GB. 
This is a problem for big Hamiltonian and Overlap matrices. We should move to another matrix format, such as hdf5.

maurodossena:
This pull request introduces support for reading and writing HDF5 (`.h5`) files for matrix data in the codebase, replacing previous usage of MATLAB `.mat` files in several examples and utilities. `.mat`files have limited size of 2GB. The main changes include adding new utility functions for handling HDF5 files, updating distributed loading logic to support `.h5` files, and updating example input files to use the HDF5 format instead of `.mat`.

This PR closes issue #311


maurodossena:
cscs-ci run

maurodossena:
cscs-ci run

maurodossena:
cscs-ci run

vetschn (approved):
Looks very good! One thing i was surprised by was that the tests passed on alps without modifying the Dockerfile, so i checked the "build quatrex" stage and it looks like there it now not only installs quatrex but also pulls scipy (!) and h5py from PyPI 💀 

I'll add those to the base Dockerfile and also a --no-deps flag in the quatrex build stage, so we'll always notice in the future.

vetschn:
cscs-ci run

vetschn (approved):

Issue #309, opened 2026-06-02
Fix variable types in the inplace rawkernels


vetschn:
> Why do we mix ints and longs and not just use clear types (u)int32/64\_t? 

 _Originally posted by @alexnick83 in [#284](https://github.com/quatrex/quatrex/pull/284/changes#r3305498515)_

almaeder:
Similarly: #78

almaeder:
Further, tests should be added for these functions.

almaeder:
Fixed by #320.

PR #320, opened 2026-06-26
Index datatype issue

### LINKED ISSUE #209: Replace RawKernels with RawModules
We should use RawModules instead of RawKernels:
- Slower compilation (single compilation call per kernel)
- Templating not possible https://github.com/cupy/cupy/pull/3319

almaeder:
This PR will address #309.

- [x] Template inplace kernels.
- [x] Move data structure kernels to rawmodule.
- [x] Add test for raw data structure kernels to the CI.
- [x] Template datastructure kernels.


almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

.cscs/alps-ci.yaml
```diff
@@ -51,9 +51,18 @@ test quatrex single:
...
+    QTX_USE_CUPY_JIT: '1'
   script:
     - coverage run --source=src --parallel-mode -m pytest -vv tests/qttools
 
+test qttools kernels:
+  extends: .test-common
+  variables:
+    QTX_USE_CUPY_JIT: '0'
+  script:
+    - coverage run --source=src --parallel-mode -m pytest -vv tests/qttools/kernels/datastructure
```
vetschn, line 64: Forgot again, JIT flag only applies to data structure kernels?

src/qttools/comm/comm.py
```diff
@@ -527,9 +527,9 @@ def all_gather_v(
                 )
             global_size = mask.size
         else:
-            counts = np.zeros(self.size, dtype=xp.int32)
+            counts = np.zeros(self.size, dtype=xp.int64)
             self.all_gather(
-                np.array(sendbuf.shape[axis], dtype=xp.int32),
+                np.array(sendbuf.shape[axis], dtype=xp.int64),
```
vetschn, line 532: Did you run into a problem with 32-bit ints? In any case a good idea 👍

src/qttools/datastructures/dsdbcoo.py
```diff
@@ -65,17 +65,29 @@ def __init__(
         symmetry_op: Callable = xp.conj,
     ):
         """Initializes a DSDBCOO matrix."""
+
+        # NOTE: The idea is that outside scipy decides
+        # the correct data type for the indices.
+        # We just need to make sure to follow.
+        index_type = rows.dtype
```
vetschn, line 72: I see, scipy seems to call a `get_index_dtype` function at instantiation that checks which dtype is applicable 👍

src/qttools/kernels/datastructure/cupy/_cupy_jit.py
```diff
@@ -1,76 +1,87 @@
...
-            my_count += cond
-        jit.syncthreads()
-
-    if i < self_rows.shape[0]:
-        full_inds[i] = my_full_ind
-        counts[i] = my_count
+_find_inds_kernels = {}
+
+
+def _find_inds(grid: tuple[int, int, int], block: tuple[int, int, int], args: tuple):
```
vetschn: You could consider doing this with a factory function `_get_find_inds_kernel` that you wrap with `functools.cache` instead of manually handling the caching logic 🤷

src/qttools/kernels/inplace/cupy/_rawkernel.py
```diff
@@ -1,75 +1,94 @@
...
+from itertools import product
+
 import cupy as cp
 
-_scatter_add_scaled_cuda_code = r"""
-#include <cupy/complex.cuh>
+value_types = [
+    (cp.float64, "double"),
+    (cp.complex128, "complex<double>"),
+]
```
vetschn: In the other `_rawkernel` file you use a dict for something almost identical i think. Maybe could be done here as well? Or have a `NamedTuple` or a tiny dataclass or something. Or have a map from cupy dtype to the other type as well and make `signatures` a dict from dtype to function name?

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
```diff
@@ -54,235 +84,308 @@ def reduction(
...
-                    && (cols[tid] >= col_start)
-                    && (cols[tid] < col_stop)
-                );
-            }
+    IndexType my_row = (i < num_self_rows) ? self_rows[i] : -1;
+    IndexType my_col = (i < num_self_rows) ? self_cols[i] : -1;
+                        
+    IndexType my_full_ind = 0;
+    IndexType my_count = 0;
+
```
vetschn: Not a real problem but there's some random whitespace in these strings fyi

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
```diff
@@ -54,235 +84,308 @@ def reduction(
...
+            IndexType cond = nnz_section_offsets[j] <= inds[i];
+            ranks[i] = ranks[i] * (1 - cond) + j * cond;
+        }
+    }
+}
+
+"""
+
+
+value_types = {
```
vetschn: you could put these dtype maps at the top level of `qttools.kernels`

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
```diff
@@ -1,51 +1,81 @@
...
 def reduction(
     a: NDArray,
-):
+) -> NDArray:
+    """Performs a reduction operation on the input array.
 
-    n_blocks = 4
+    NOTE: This is a naive implementation for SC25
+    This was needed since cupy didnt perform well on MI250X.
+    TODO: Further investigate on newer cupy versions.
```
vetschn: You can have a "Notes\n-----" section

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
```diff
@@ -54,235 +84,308 @@ def reduction(
...
-                                         
-            if (i < num_self_rows) {{
-                full_inds[i] = my_full_ind;
-                counts[i] = my_count;
-            }}
-        }}
-    """,
-    "find_inds",
-)
+kernels_template = r"""
```
vetschn: How cooked do you think it would be to have these strings in actual `.cu` files and load them at the module level?

src/qttools/kernels/datastructure/cupy/dsdbcsr.py
```diff
@@ -78,28 +85,49 @@ def find_inds(
         rowptr = rowptr_map.get((brow, bcol), None)
         if rowptr is None:
             continue
-        mask = cp.zeros(brows.shape[0], dtype=cp.bool_)
+        mask = cp.zeros_like(brows)
```
vetschn: Can you quickly explain why this is no longer a boolean?

tests/qttools/kernels/inplace/test_inplace.py
```diff
@@ -0,0 +1,69 @@
...
+    a = sparse.random(*shape, density=0.2, format="csr") + 1j * sparse.random(
+        *shape, density=0.2, format="csr"
+    )
+    b = sparse.random(*shape, density=0.2, format="csr") + 1j * sparse.random(
+        *shape, density=0.2, format="csr"
+    )
+
+    # make a superset of the indices of b in a
+    a = (a + b).tocoo()
+    b = b.tocoo()
```
vetschn, line 34: Why not `coo` from the start?

src/qttools/kernels/inplace/cupy/inplace.py
```diff
@@ -31,13 +35,24 @@ def scatter_add_scaled(
...
-    kernel = _rawkernel.scatter_add_scaled_kernels.get(
-        (a.dtype.type, b.dtype.type, alpha.dtype.type)
-    )
+    if isinstance(alpha, complex):
+        alpha_type = cp.complex128
+    elif isinstance(alpha, float):
+        alpha_type = cp.float64
+    else:
+        raise TypeError(
+            f"Unsupported type for alpha: {type(alpha)}. Must be float, complex, or NDArray."
```
vetschn: What do you mean by "or NDArray?"

vetschn (approved):
I left a few comments but it's looking quite good overall 👍

.cscs/alps-ci.yaml
almaeder: yes.
Notes:
JIT is the default.
The `inplace` kernels for QTBM do not have a JIT version.

src/qttools/comm/comm.py
almaeder: Not yet since none of the structures are that big for GW.
I was searching where we hardcode `int32` and `int64` and it seemed kinda dangerous here to have `int32`,
Here there should be no performance problem to default to int64 for a single value.
Side note: https://github.com/mpi4py/mpi4py/discussions/300

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
almaeder: ☠️

src/qttools/kernels/inplace/cupy/inplace.py
almaeder: Ah, message was not updated.
NDArray was wrongly allowed while the kernel only takes a scalar.

src/qttools/kernels/datastructure/cupy/dsdbcsr.py
almaeder: Only a bit laziness.  😴

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
almaeder: It would be nicer for formatting. 
I remember discussing this, but it never materialising.
Would it then be fine to include `clang-format` as a dependency?

tests/qttools/kernels/inplace/test_inplace.py
almaeder: Will cast to `csr` during the addition and then one needs to cast back.

almaeder:
cscs-ci run

almaeder:
cscs-ci run

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
```diff
@@ -1,288 +1,155 @@
...
+            f"_reduction<{idx1[1]}, {idx2[1]}>"
+        )
+
+for idx in index_types.items():
+    name_expressions[cp.bool_, idx[0], "_reduction"] = f"_reduction<bool, {idx[1]}>"
+
+for idx in index_types.items():
+    for val in value_types.items():
+        name_expressions[(idx[0], val[0], "_densify_block")] = (
+            f"_densify_block<{idx[1]}, {val[1]}>"
```
vetschn: Why not just unpack the keys and values 💀

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
```diff
@@ -1,288 +1,155 @@
...
+)
 
-def reduction(
-    a: NDArray,
+kernels = {key: module.get_function(value) for key, value in name_expressions.items()}
+
+
+def _find_inds(grid: tuple[int, int, int], block: tuple[int, int, int], args: tuple):
+    kernel = kernels[(args[0].dtype.type, "_find_inds")]
+    kernel(grid, block, args)
```
vetschn, line 68: It's actually fine but you like it is but you already have a list of the kernel names and the definitions are all almost identical, so looks like the duplication can easily be reduced 🤷

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
almaeder: Would it be clearer? Then we would have something like `idx_numpy_type`, `idx_c_type`, `idx1_numpy_type`, `idx1_c_type`, `idx2_numpy_type`, `idx2_c_type`.

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
almaeder: The only problem is if the type inference is not from the first argument. 
We could have another descriptor describing from which arguments to infer the types.

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
vetschn: To me yes, but it's not an actual problem

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
```diff
@@ -1,288 +1,155 @@
...
-    """,
-    "_find_ranks_kernel",
-)
+    n_blocks = 4
+    out = cp.zeros((n_blocks * THREADS_PER_BLOCK), dtype=dtype)
+
+    n = a.size
+
+    _reduction = kernels[(index_types[dtype], "_reduction")]
+
```
almaeder: There seems to be a bug in the tests because this is wrong, but the tests do not fail...

src/qttools/kernels/datastructure/cupy/_cupy_rawkernel.py
almaeder: But I would accept here the boilerplate code.
A descriptor to auto-generate the functions seems a bit much.
Would leave it for now, and if we ever have to add more functions, it can be discussed again.

almaeder:
cscs-ci run

PR #325, opened 2026-07-01
Toggle to enable saving of intermediate SCBA variables and visualization script

YongDaLi:
This is the code and visualization jupyter notebook that I used for my thesis to gather and dump the SCBA variables during each SCBA loop. There's 2 parameters in the config. One for enabling saving and another for specifying how many of the nnz index to save (it would generate too much data to save all nnz). 

The visualization notebook uses a `scba_container` class to help organize the data and plotting. There's a magic number 43823 that's hard-coded for the w90/carbon-nanotube/gw example. There's 43824 nnz indices in that example. I thought about making it automatically detect the number of nnz indices, but I'm not sure how the data is organized in the other examples. I'm only familiar with this example.


YongDaLi:
Talked with @vetschn. The data for the other examples are structured similarly, so it should be able to autodetect the size of the nnz dimension. Additionally you should be able to save all data (don't worry about storage space. Saving TB's of data is fine, especially on Alps). This is more of a debugging hacky thing, so maybe it should not be in production code.

I'll add these 2 features anyway. 
1. auto-detect nnz dimension size
2. toggle to save all nnz samples

YongDaLi:
Not working for k-points since the `self.data.g_lesser.shape` changes from ex. (10 iterations, 768 atoms, 768 atoms) to (10 iteraitons, 3 k-points, 768 atoms, 768 atoms). Messes up my mpi gather wrapper `gather_array_stack` since it does not expect the k-point dimension, so the gatherbuf is too small.

YongDaLi:
image
It would be nice if it could automatically detect the k-point dimension and use the appropriate mpi gather wrapper. Then it can print the row, col, and k-point all in the plot title.

PR #327, opened 2026-07-07
Fix batched RGF currents and add dense bond current

### LINKED ISSUE #67: De-coupling further of RGF and linear solver
<a href="https://github.com/vincent-maillou"></img></a> **Issue by [vincent-maillou](https://github.com/vincent-maillou)**
_Friday Feb 14, 2025 at 11:14 GMT_
_Originally opened as https://github.com/vincent-maillou/qttools/issues/90_

----

At the end, aren't the green’s function solver.. Solvers for the green's function? Meaning that yes there is a physical embedding at some point, may it be OBC or Meir-Wingreen current computation.
We could decouple the linear solver part of GFSolver further to be able to "plug" a novel linear solver more easily.

For example:
```
def greens_function_solver(A, Sigma):
    g = linear_solver.schur(A, Sigma)
    mw_current = compute_MW(g)
    G = linear_solver.sellinv(A, g, Sigma)
    return mw_current
```

Kosheira1:
RGF current bug fixes (rgf.py):
- Slice OBC blocks and the current assignment with the batch stack_slice in the contact-current computation; fixes a broadcast error when max_batch_size < num_energies.
- Move the contact-current block inside the batch loop; it previously ran only for the last batch, leaving other batches' contact currents zero. Interior currents were unaffected.

Dense bond current (device_current_dense):
- Add an optional coherent bond current computed from the dense off-diagonal G< blocks inside the RGF, before they are truncated onto the interaction-cutoff sparsity pattern. Gated by outputs.device_current_dense; matches the Meir-Wingreen current for long-range, non-orthogonal systems. Not supported with block parallelism.

device_current (observables.py):
- Optionally subtract the energy-weighted overlap (H - E*S) for non-orthogonal bases; falls back to H-only when no overlap is given.


src/qttools/greens_function_solver/rgf.py
```diff
@@ -411,28 +451,35 @@ def selected_solve(
...
-            )
-            # NOTE: Negative sign is needed to get the current flowing
-            # in the correct direction (positive from left to right).
-            current[..., -1] = -xp.trace(
-                obc_blocks.greater[-1] @ xl_diag_blocks[-1]
-                - xg_diag_blocks[-1] @ obc_blocks.lesser[-1],
-                axis1=-2,
-                axis2=-1,
-            )
+            # NOTE: This must stay inside the batch loop so it uses the
```
almaeder: Nice catch 🔥

src/quatrex/core/config.py
```diff
@@ -1045,6 +1045,12 @@ class OutputConfig(BaseModel):
     # Only the spectral currents are saved by default.
     device_currents: bool = True
 
+    # Coherent bond current between adjacent blocks, computed from the
+    # dense (untruncated) off-diagonal Green's function blocks inside the
+    # RGF solver. Untruncated analogue of the `device` current. Only
+    # supported with the `rgf` solver and without block parallelism.
+    device_current_dense: bool = False
```
almaeder: Maybe we can make a validator to check when we are in the domain distributed case.
Then we would stop earlier before the `raise` in `rgf_dist`.

src/quatrex/electron/solver.py
```diff
@@ -847,31 +858,24 @@ def solve(
...
+                result = solver.selected_solve(
+                    a=self.system_matrix,
+                    sigma_lesser=sse_lesser_batch,
+                    sigma_greater=sse_greater_batch,
+                    obc_blocks=self.obc_blocks,
+                    out=out_slice,
+                    return_retarded=True,
+                    return_current=self.compute_meir_wingreen_current,
+                    return_bond_current=self.compute_bond_current,
+                )
```
almaeder, line 1095: 👍

src/quatrex/core/scba.py
```diff
@@ -612,6 +624,10 @@ def _write_iteration_outputs(self, iteration: int):
                 outputs[f"meir_wingreen_current_{iteration}.npy"] = (
                     self.observables.electron_current["meir-wingreen"]
                 )
+            if self.config.outputs.device_current_dense:
+                outputs[f"device_current_dense_{iteration}.npy"] = (
+                    self.observables.electron_current["device-dense"]
+                )
```
almaeder: Not the biggest fan of the name with `_dense`, but unsure what a better name would be to differentiate both currents.

almaeder (commented):
Nice work. Looks already very good.

almaeder:
Edit: Alps is currently down and will be tested later.

src/qttools/greens_function_solver/rgf.py
alexnick83: We should consider slightly optimizing the trace computations in a future PR, since we only need the main diagonals of the product. Likely something similar to `xp.einsum('kij, kji -> k')` would work.

src/qttools/greens_function_solver/rgf_dist.py
```diff
@@ -169,8 +170,17 @@ def selected_solve(
             the Meir-Wingreen formula. By default False. Note that this
             is currently only partially supported, and only the boundary
             currents are computed correctly.
+        return_bond_current : bool, optional
+            Not supported in the block-distributed solver. By default
+            False; setting it True raises `NotImplementedError`.
 
         """
+        if return_bond_current:
+            raise NotImplementedError(
```
alexnick83: If I understand the formula correctly, we can compute this with the distributed RGF, since it depends on the system matrix and the (final) G lesser, unlike the Meier-Wingreen. Do we want to support this (perhaps through a future PR)?

src/quatrex/core/scba.py
alexnick83: Why not just call it `bond_current` or `coherent_bond_current`?

almaeder: @vetschn was suggesting `liouville_current` since also `device` is not saying a lot.
Maybe `device_current` -> `liouville_current_x`  and `device_current_dense` ->  `liouville_current_y` with x/y being `_approx`/ `_exact`, `_sparse`/`_dense`, or `_sparse`/`_full`?
I would choose `_approx`/ `_exact.

almaeder: Ah, just delete the original approximated current and implement the correct one also with `rgf_dist`.

alexnick83:
cscs-ci run

almaeder:
I will try to push this PR and make further refactors.

almaeder:
cscs-ci run

almaeder:
Note: After discussing with Mathieu, for the new current in RGF, we are not allowed to use `a` since it contains the self-energy. We have to pass in a separate `a_hat`, but I wouldn't allocate new memory for it, but just have a shell that allows views and block access and materialize the blocks lazily. 

Similarly, as with the old code, we can do the same for `a`.  I do not believe the overhead of the extra additions is problematic.

~~Further, the code should be only merged after verifying the results against QTBM.~~ Done.

Note:
This will be done in a convoluted way, but only after #333 is merged. There, I refactor the stack view class to make it easier to build on top of it and pass in a custom block indexer.

vetschn:
cscs-ci run

vetschn:
@almaeder already made some very nice improvements here with the `SystemMatrix` class. We just discussed irl that we should also change the computation of the charge density in systems with non-orthogonal overlap to be done on the dense blocks instead. To try to avoid making this even more specific to the electronic part and passing more stuff into the solver, i'll look into doing this with callbacks during on the back-substitution pass.

Another thing to be discussed is whether the commutator between the bare Hamiltonian and the density matrix actually gives the full picture of the current. I think it's possible that we should instead be including scattering terms in a sort of collision integral like in the [Lindblad master equation](https://en.wikipedia.org/wiki/Lindbladian). Unsure and to be discussed

vetschn:
cscs-ci run

vetschn:
cscs-ci run

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -855,9 +855,8 @@ def _replace_ellipsis(stack_index: tuple, ndim: int) -> tuple:
             raise IndexError("an index can only have a single ellipsis ('...')")
 
         loc = is_ellipsis[0]
-        extra_dimensions = (ndim - 1) - (
-            len(stack_index) - sum(i is None for i in stack_index) - 1
-        )
+        explicit_indices = len(stack_index) - sum(i is None for i in stack_index) - 1
+        extra_dimensions = ndim - explicit_indices
```
almaeder, line 859: I changed this since it was a bug that wasn't noticed.
It emerged when doing the system matrix class.

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -1105,6 +1106,11 @@ def _make_block_indexer(self) -> "_DSDBlockIndexer":
             dsdbsparse=self._dsdbsparse, stack_index=self._stack_index
         )
 
+    @property
+    def block_section_offsets(self) -> list[int]:
+        """Returns the block section offsets."""
+        return self._dsdbsparse.block_section_offsets
+
```
almaeder: Shouldn't we add this to `_DELEGATED`?

src/qttools/greens_function_solver/_serinv.py
```diff
@@ -693,6 +699,20 @@ def scatter(
                 is_retarded=False,
             )
 
+            # NOTE: This is done here because it is the last time we
+            # have access to the dense off-diagonal blocks connecting
+            # partitions, since they are sparsified during the mapback.
+            if callbacks is not None and comm.block.rank != comm.block.size - 1:
```
almaeder, line 705: Why do we need to check that not the last block doesn't call it?

src/qttools/greens_function_solver/_serinv.py
```diff
@@ -113,13 +113,14 @@ def gather(
...
+        xl_buffer_upper: list[NDArray] | None = None,
+        xl_buffer_lower: list[NDArray] | None = None,
         sigma_greater: DSDBSparse | _DStackView = None,
-        xg_diag_blocks: list[NDArray] = None,
-        xg_buffer_upper: list[NDArray] = None,
-        xg_buffer_lower: list[NDArray] = None,
+        xg_diag_blocks: list[NDArray] | None = None,
+        xg_buffer_upper: list[NDArray] | None = None,
+        xg_buffer_lower: list[NDArray] | None = None,
+        **kwargs,
```
almaeder, line 123: I guess the `kwargs` does make the interface nicer for calling the different functions of `_serinv`?

vetschn: yep, that's the point of the kwargs, i didn't want to convert the whole thing to a dataclass, i thought it may make it even worse to read

src/qttools/datastructures/dsdbsparse.py
vetschn: yes

src/qttools/greens_function_solver/_serinv.py
vetschn: These separator blocks that are needed to compute the current between partitions only exist on the partitions that connect to another subdomain on their lower right. Last partition does not have the fishtail on the lower right, so no "device" current flow there. Does that make sense?

almaeder: Ah yes, makes sense. Thanks 🔥 
Some questions are also to remind myself of the changes.

vetschn:
Waiting for alps to be back to investigate the failing test on the electron_ldos_0 of the gw-dist reference test on daint. At iteration 1 the ldos matches well there, but at 0 it is slightly above the tolerance. The reference data was generated with `"full"` OBC algo (but on CPU) so i guess this is a different source of noise than the Beyn RNG?

vetschn:
Was able to check now on alps, and it really looks like a tiny noise issue, that causes the `np.allclose` to fail for the `electron_ldos_0.npy`. So besides the device current files, there is now also a new ldos reference output file for the gw-dist example.

Here is the _absolute_ difference between actual output and reference (just above `1e-4`):
```python
fig, ax = plt.subplots()

qm = ax.pcolormesh(np.abs(output - reference), norm="log")
fig.colorbar(qm, ax=ax, label="|output - reference|")
```
image

vetschn:
cscs-ci run

vetschn (approved):

almaeder (approved):
Thank you pulling it @vetschn

almaeder:
I guess both the old and new current look both constant?
We always had problems with this example since I believe the old reference data was still generated when using beyn.

vetschn:
> We always had problems with this example since I believe the old reference data was still generated when using beyn.

I actually checked the commit the previous data was added, and the corresponding config was actually using full instead. I'm not quite sure where the problem came from.

almaeder:
It can still be that the operation order changed, and this specific config is unstable.

Issue #83, opened 2025-09-05
Symmetry string instead of symmetry flag + op


almaeder:
<a href="https://github.com/vetschn"><img src="https://avatars.githubusercontent.com/u/63104434?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [vetschn](https://github.com/vetschn)**
_Monday May 19, 2025 at 13:08 GMT_
_Originally opened as https://github.com/vincent-maillou/qttools/issues/152_

----

Instead of passing a function we should have keywords like `"symmetric"`/`"hermitian"`/`"skew-hermitian"` etc. and look that up in a `dict`. The `.symmetrize()` method should also be amended

almaeder:
This was done with #328.

Issue #223, opened 2026-02-03
Block sizes setting functionality


almaeder:
Currently, the block size change induces recomputation of diagonal indices. Ideally, these would be cashed.

- [ ] Have a look at the `block_size` setter method, this could use some work
- [ ] Implement a caching mechanism

almaeder:
Implemented in #333.

PR #328, opened 2026-07-09
`DSDBSparse` allocation interface refactor

almaeder:
This PR will refactor the allocation interface of our matrices and refactor other data structure parts:

- [x] Not defaulting to allocate memory for the data vector.
- [x] Remove unused functionality.
- [x] Fixing issue #83.

This PR will touch too many files and will be a bit too big 💀.


almaeder:
@vetschn if you have time could you already have a look?
I am kinda happy with the changes. 
I am pondering since it is, anyway, too many changes if I should also just include #223.

almaeder:
cscs-ci run

src/quatrex/core/scba.py
```diff
@@ -740,6 +759,8 @@ def run(self) -> None:
             with profiler.profile_range(
                 label="SCBA: Iteration", level="default", comm=comm
             ):
+                self.data.g_retarded.allocate_data()
+                self.data.g_retarded.data = 0.0
```
almaeder, line 763: The zero setting is here only to make the phonon test not fail: the phonon test is currently broken since the off-diagonals of Gr are not computed.
See #332.

almaeder:
cscs-ci run

almaeder:
cscs-ci run

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -118,92 +126,124 @@ def __init__(
...
-        symmetry_op: Callable = xp.conj,
+        symmetry: str | None = None,
     ):
         """Initializes a DSBDSparse matrix."""
 
+        # Set the block and stack communicators.
+        if comm.block is None or comm.stack is None:
+            raise ValueError(
+                "Block and stack communicators must be initialized via "
+                "the BLOCK_COMM_SIZE environment variable."
```
vetschn: They can also be set via the config, right?

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -899,23 +835,39 @@ def allocate_data(self, stack_size: int | None = None) -> None:
...
+        allocate: bool = True,
     ) -> "DSDBSparse":
         """Creates a new DSDBSparse matrix from a scipy.sparse array.
 
+        This essentially distributed the matrix across the stack and
+        block communicators.
+
+        !!! Note
+            There is no data allocated in the new matrix and only the
+            sparsity pattern is used.
```
vetschn: I guess this is outdated with the allocate parameter?

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -13,6 +13,14 @@
 from qttools.utils.gpu_utils import free_mempool, synchronize_device
 from qttools.utils.mpi_utils import get_section_sizes
 
+symmetry_ops = {
+    None: lambda x: x,
```
vetschn: I expect that `None` is (and should) never be requested from this dict, right? Maybe remove to force a `KeyError` if this should happen anyway?

src/quatrex/core/scba.py
```diff
@@ -177,6 +177,16 @@ def __init__(self, config: QuatrexConfig, electron_energies: NDArray) -> None:
...
 
+        # Allocate the data for the Green's functions and self-energies.
+        self.g_lesser.allocate_data()
+        self.g_greater.allocate_data()
+        self.sigma_lesser.allocate_data()
+        self.sigma_greater.allocate_data()
+        self.sigma_retarded_hermitian.allocate_data()
+        self.sigma_lesser.data = 0.0
+        self.sigma_greater.data = 0.0
+        self.sigma_retarded_hermitian.data = 0.0
```
vetschn, line 188: I guess these are allocated in the `__init__` because they persist throughout a run, right? Maybe could be revisited when looking at lifetimes of different quantities.

vetschn (approved):
Looks very good! Besides the few comments i left, you could maybe just add one sentence to the `allocate_data` docstring to explicitly say that the data is _not initialized_

src/qttools/datastructures/dsdbsparse.py
almaeder: This docstring is wrong, and I copied it around. There is no `BLOCK_COMM_SIZE` variable anymore.

src/qttools/datastructures/dsdbsparse.py
almaeder: Ah true. Good catch :fire:.

src/quatrex/core/scba.py
almaeder: Both allocating them here and allocating them at the start of `run` would be fine.
This probably needs to be revisited in the context of electrostatics.

It will be revisited when I try to push the number of energies again.

vetschn: Why exactly in the context of electrostatics?

almaeder: Since this leads to another out loop, right?
Just there to ensure we don't deallocate things that will be needed for the next loop iterations.
I currently think it should work.

almaeder:
cscs-ci run

vetschn:
cscs-ci run

Issue #224, opened 2026-02-03
Rethink DSDBSparse Views


almaeder:
The block and stack views contain some logic that stems from the original data layout with padding in the data buffer. This can probably be simplified.

There is also logic regarding setting and getting items in stack views. We may also need to perform arithmetic on these views. We'll have to revisit the design here.

almaeder:
Refactored in #333.

Issue #82, opened 2025-09-05
The diag_inds are not updated when changing block sizes


almaeder:
<a href="https://github.com/awinka"><img src="https://avatars.githubusercontent.com/u/131971675?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [awinka](https://github.com/awinka)**
_Thursday Apr 03, 2025 at 08:09 GMT_
_Originally opened as https://github.com/vincent-maillou/qttools/issues/149_

----

When initializing a `dsbcsr` object  the properties `self._diag_inds` and `self._diag_value_inds` are calculated from the `self.rows` and `self.cols`. When changing block sizes the `self.rows` and `self.cols` are permuted, but the `self._diag_inds` and `self._diag_value_inds` stays the same. This means that the `diagonal()` method will give wrong results after the change of block sizes.

almaeder:
Fixed in #333.

PR #333, opened 2026-07-13
Refactoring setting the block size

almaeder:
This PR addresses issues #82, #223, #224, and a subpart of #83.
The issues are all related to refactoring the data structure.

Note:
#83 was a leftover that was not fixed in #328.

Note:
Removes more unused functionality that should not be used.

The view interface was refactored to later be used for the system matrix.


almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

vetschn:
cscs-ci run

examples/w90/carbon-nanotube/gw-dist/quatrex_config.toml
vetschn, line 1: Changes in these files now also lead to a proper test of the nested batching, right?

src/qttools/datastructures/dsdbsparse.py
vetschn:
```suggestion
class BlockConfig:
```

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -293,14 +263,25 @@ def __init__(
...
+        # the mapping from one block configuration to another.
+        # NOTE: Currently, it is assumed that each configuration
+        # is uniquely identified by the number of blocks. This is
+        # not necessarily true, but it is a reasonable assumption for now.
+        self._block_change_cache: dict[(int, int), NDArray] = {}
         self._add_block_config(self.num_blocks, block_sizes, block_offsets)
 
         self._block_indexer = _DSDBlockIndexer(self)
-        self._stack_indexer = _DStackIndexer(self)
+        self._stack_indexer = _DStackView(
```
vetschn: Consider renaming the attribute to to `_stack_view`

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -739,22 +655,18 @@ def spy(self) -> tuple[NDArray, NDArray]:
         ...
 
     @abstractmethod
-    def symmetrize(self, op: Callable[[NDArray, NDArray], NDArray] = xp.add) -> None:
-        """Symmetrizes the matrix with a given operation.
-
-        This is done by setting the data to the result of the operation
-        applied to the data and its conjugate transpose.
+    def symmetrize(self, symmetry: str) -> None:
```
vetschn, line 658: I guess this got overlooked in the recent PR concerning symmetries?

src/qttools/datastructures/dsdbsparse.py
```diff
@@ -1018,127 +934,158 @@ def ensure_tuple(ndslice):
...
 
-        stack_index = _replace_ellipsis(stack_index, self._dsdbsparse.data.ndim)
-
-        if sparse.issparse(other):
-            csr = other.tocsr()
-            self._dsdbsparse.data[stack_index] = csr[self._dsdbsparse.spy()]
-            return None
-
-        self._dsdbsparse.data[stack_index] = other.data[stack_index]
+    Supports further indexing via `[...]` or `.stack[...]`, which
```
vetschn, line 973: If i understand correctly, this simplifies a lot the logic that was done before to achieve stuff like this for batching?

```python
a.stack[1:5].stack[:3].stack[2:]
```

src/qttools/datastructures/dsdbcoo.py
```diff
@@ -115,188 +101,40 @@ def _set_diagonal_indices(self) -> None:
...
-        self._diag_inds_nnz = (
-            self._diag_inds[ranks == comm.stack.rank]
-            - self.nnz_section_offsets[comm.stack.rank]
-        )
-        self._diag_value_inds_nnz = (
-            self._diag_value_inds[ranks == comm.stack.rank]
-            - self._diag_value_inds[ranks == comm.stack.rank][0]
-        )
-
-    def _get_items(self, stack_index: tuple, rows: NDArray, cols: NDArray) -> NDArray:
```
vetschn, line 131: It looks like removing the item getters/setters also makes the `find_inds` kernels unused (except for the tests 😅 ).

🪚 ?

src/qttools/datastructures/dsdbcoo.py
```diff
@@ -484,138 +296,67 @@ def block_sizes(self, block_sizes: NDArray) -> None:
...
         if num_blocks in self._block_config and num_blocks == self.num_blocks:
             return
 
         if sum(block_sizes) != self.shape[-1]:
             raise ValueError("Block sizes must sum to matrix shape.")
 
-        block_section_sizes, __ = get_section_sizes(len(block_sizes), comm.block.size)
-        block_section_offsets = np.hstack(([0], np.cumsum(block_section_sizes)))
+        if (self.num_blocks, num_blocks) in self._block_change_cache:
+            inds_bcoo2bcoo = self._block_change_cache[self.num_blocks, num_blocks]
```
vetschn, line 325: Thanks, way better than the confusing, duplicated if/else branching that was there before

src/qttools/datastructures/dsdbcoo.py
```diff
@@ -484,138 +296,67 @@ def block_sizes(self, block_sizes: NDArray) -> None:
...
-        inds_canonical2bcoo = dsdbcoo_kernels.compute_block_sort_index(
-            canonical_rows, canonical_cols, local_block_sizes
-        )
-        # Mapping directly from original block-ordering to the new
-        # block-ordering is achieved by chaining the two mappings.
-        inds_bcoo2bcoo = inds_bcoo2canonical[inds_canonical2bcoo]
-        data = self.data.reshape(-1, self.data.shape[-1])
-        for stack_idx in range(data.shape[0]):
-            data[stack_idx] = data[stack_idx, inds_bcoo2bcoo]
+        self.data = self.data[..., inds_bcoo2bcoo]
```
vetschn: Since we are fancy-indexing here, this can end up copying a huuge amount of data if done this way directly, which is why there was a for loop around this previously.

I would argue for keeping at least a batching loop around this, it does not cost a lot of time and won't cause memory spikes that we'll eventually just be investigating again.

Definitely worth adding a `# NOTE:` about this

src/qttools/datastructures/dsdbcsr.py
```diff
@@ -115,174 +109,7 @@ def __init__(
             - self._diag_value_inds[ranks == comm.rank][0]
         )
 
-    def _get_items(self, stack_index: tuple, rows: NDArray, cols: NDArray) -> NDArray:
```
vetschn, line 118: For the CSR case the find_inds kernel is still used to determine the diagonal indices, so unlike for the COO i don't think it can be removed here

src/qttools/datastructures/dsdbcsr.py
```diff
@@ -485,31 +266,49 @@ def block_sizes(self, block_sizes: NDArray) -> None:
...
         # block-ordering is achieved by chaining the two mappings.
         inds_bcsr2bcsr = inds_bcsr2canonical[inds_canonical2bcsr]
-        data = self.data.reshape(-1, self.data.shape[-1])
-        for stack_idx in range(data.shape[0]):
-            data[stack_idx] = data[stack_idx, inds_bcsr2bcsr]
+        block_offsets = np.hstack(([0], np.cumsum(block_sizes)))
+
+        self._add_block_config(num_blocks, block_sizes, block_offsets)
+
+        self.data = self.data[..., inds_bcsr2bcsr]
```
vetschn: See comment for the COO case

src/qttools/greens_function_solver/rgf_dist.py
```diff
@@ -240,61 +213,122 @@ def selected_solve(
                 )
 
             batch_sizes, batch_offsets = get_batches(
-                sigma_lesser.shape[0], self.max_batch_size
+                sigma_lesser.local_stack_shape[0], self.max_batch_size
             )
 
-            for i in range(len(batch_sizes)):
-                stack_slice = slice(int(batch_offsets[i]), int(batch_offsets[i + 1]))
+        for i in range(len(batch_sizes)):
```
vetschn, line 219: What is the benefit of having these quantities initialized in each batching loop? I'm probably missing something but the list elements get overwritten for every batch anyway, no?

src/qttools/greens_function_solver/rgf_dist.py
```diff
@@ -240,61 +213,122 @@ def selected_solve(
...
+                None
+            ] * sigma_lesser.num_local_blocks
 
-        with profiler.profile_range(
-            label="RGF dist: Schur", level="default", comm=comm
-        ):
+            xl_diag_blocks: list[NDArray | None] = [
+                None
+            ] * sigma_lesser.num_local_blocks
+            xl_buffer_lower = None
```
vetschn, line 237: Does the diff just look weird or what was going on here before? Was it not actually doing the batching? If yes, for how long has this been wrong?

vetschn (commented):
Block change cache is a nice solution 👍 Stack indexing also looks great, no more indirection via this indexer. Also i guess a first step to untangling the DSDBSparse into block-access and stack-access components

src/qttools/greens_function_solver/rgf_dist.py
almaeder: :skull: No batching was done before. This was wrong on old `qttools` already,
but this functionality was never used in practice since everyone only had one energy.
If you have to use the distributed solver, then there should be only one energy. I can not think of an edge case where you would batch in practice.

src/qttools/greens_function_solver/rgf_dist.py
almaeder: There is no good reason except that it is clearer to me that things do not carry through the loops.

examples/w90/carbon-nanotube/gw-dist/quatrex_config.toml
almaeder: Yes, since both batch sizes are set and smaller than the total number of energies per rank.

src/qttools/datastructures/dsdbsparse.py
almaeder: Yes...

src/qttools/datastructures/dsdbsparse.py
almaeder: Why was this not auto-formatted?

src/qttools/greens_function_solver/rgf_dist.py
vetschn: Oops :skull: 

Agreed, but of course still better to have this working

src/qttools/datastructures/dsdbsparse.py
almaeder: Yes, before we had the two classes creating each other, while there is now just one. 
The major difference is that a view only supports minimal methods like `.blocks` and accessing certain attributes.

src/qttools/datastructures/dsdbcoo.py
almaeder: 🪓

src/qttools/datastructures/dsdbcoo.py
almaeder: Ah, I was a bit confused about why the loop was here. Will add it back with the note.

src/qttools/datastructures/dsdbcoo.py
almaeder: I am now doubting the caching of the block change indices:
Should we allow this since it can consume "a lot" of memory?

src/qttools/datastructures/dsdbsparse.py
vetschn: Ah before there was also getting individual items on a stack view :+1:

src/qttools/datastructures/dsdbcoo.py
almaeder: Caching the diagonal indices is fine since they are O(N).

src/qttools/datastructures/dsdbsparse.py
vetschn: Ah this was about inheriting from object, which is not necessary

src/qttools/datastructures/dsdbsparse.py
almaeder: There were many methods on the stack view like `__iadd__`, but I removed all of them over the last PRs.

src/qttools/greens_function_solver/rgf_dist.py
almaeder: Since the test for it used the sparsity of the outputs, this flew under the radar.

src/qttools/datastructures/dsdbcoo.py
vetschn: It will be another `nnz` integers. If this turns out to cause many problems, we can easily deactivate the cache completely.

almaeder: Fair, but it will be "hard" to detect that this memory consumption is the problematic one.
It is most problematic since block change is only used in `W` on the largest nnz quantities.
If we saved our data in `fp32` then we double the consumption in the edge case of a single energy.
To be considered: putting these indices on the host.

almaeder:
cscs-ci run

vetschn (approved):

PR #337, opened 2026-08-03
interface to PETSc's "KSP" direct and iterative solvers

vetschn:
This interface should allow us to use any of the solvers exposed via PETSc. This includes MUMPS (CPU-only), strumpack, pastix and others.

I tried making it as GPU-aware as i could. Set up of the system should not incur a lot of copied data and (in my understanding) no device/host transfers.
The distributed and sequential matrix types are automatically inferred, seems to work from a small test. Reusing analysis and factorization is a bit half-baked and untested. Different matrix types and views also still need to be tested further and so does the distributed mode.

### DIFF
--- a/src/qttools/wave_function_solver/petsc.py
+++ b/src/qttools/wave_function_solver/petsc.py
@@ -1,0 +1,409 @@
+# Copyright (c) 2024-2026 ETH Zurich and the authors of the qttools package.
+
+
+try:
+
+    import petsc4py
+    from petsc4py import PETSc as petsc
+
+    petsc_matrix_types = {
+        "real_symmetric_positive_definite": petsc.Mat.Option.SPD,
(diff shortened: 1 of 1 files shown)


vetschn:
cscs-ci run

vetschn:
cscs-ci run

vetschn:
Quick summary:

- I added the installation of PETSc/petsc4py to a pixi target for the CI. This build is ScalarType `"complex"` only. CPU tests pass with superlu_dist on non-symmetric matrices
- Added an installation of PETSc/petsc4py to the CSCS Dockerfile, again ScalarType complex but with cuda support and a bunch of additional solvers. Distributed tests pass.
- The petsc WFSolver interface itself is a bit cooked since i had to call directly into the CDLL as petsc4py does not expose the methods that would be needed to do more efficient data allocation on GPU. Still, PETSc KSP solvers do not accept device memory, so D2H transfers are inevitable with this library.

PR #340, opened 2026-08-06
Maurodossena/print charge

maurodossena:
This PR adds printing of the excess charge at the end of QTBM calculations. Both excess electrons and excess holes are printed, in two separate .npy files.
The config flag atom_resolved_outputs allows the excess charges and DOS files to be printed atom-resolved instead of orbital-resolved.
The tests were updated accordingly.

In addition, the QTBM docs were slighly refactored.


maurodossena:
cscs-ci run

docs/user_guide/methodology/qtbm.md
```diff
@@ -15,25 +15,37 @@ of the cost of a full NEGF calculation.
...
-one of the contacts, so a device with several contacts and several open
-channels per contact is solved as a single linear system with multiple
-right-hand sides. The same boundary self-energies
-$\mathbf{\Sigma}^{R}_{OBC}(E)$ discussed in the [OBC section](obc.md)
-are used here. However, QTBM always relies on the [`spectral` OBC
-algorithm](obc.md#spectral-method) to compute the boundary
-self-energies, since the injection vectors directly derive from the
-eigenpairs of the polynomial eigenvalue problem (see [NEVP](nevp.md))
-that is solved in this OBC algorithm.
+Rather than inverting the system matrix, given the electron energy $E$
```
almaeder, line 23: Nice to add kpoints 👍

src/quatrex/core/config.py
```diff
@@ -2298,17 +2302,6 @@ def check_either_fermi_or_midgap(self) -> Self:
...
-            if (
-                contact.fermi_level is not None
-                and contact.mid_gap_energy is not None
-                and not (self.electron.band_edge_tracking or self.scsp is not None)
-            ):
-                raise ValueError(
-                    "Both `fermi_level` and `mid_gap_energy` cannot be set "
-                    "simultaneously, unless band edge tracking is active "
-                    "or the Schrödinger-Poisson solver is enabled."
-                )
```
almaeder, line 2310: Could you remind me why this was removed?

src/quatrex/core/qtbm.py
```diff
@@ -1118,7 +1125,36 @@ def _write_outputs(self):
             for contact, ldos in self.observables.electron_ldos.items():
                 np.save(
                     f"{output_dir}/dos_{contact.name[0]}.npy",
-                    ldos,
+                    (
+                        np.add.reduceat(ldos, self.device.orbital_offsets[:-1], axis=1)
```
almaeder: Should the reduction be on the host?

src/quatrex/core/qtbm.py
```diff
@@ -1164,12 +1200,17 @@ def _compute_excess_charge_densities(self):
...
-        excess_hole_density = xp.trapezoid(hole_density, self.electron_energies, axis=1)
+        self.observables.excess_hole_density = xp.trapezoid(
+            hole_density, self.electron_energies, axis=1
+        )
 
-        return excess_electron_density, excess_hole_density
+        return (
+            self.observables.excess_electron_density,
+            self.observables.excess_hole_density,
+        )
```
almaeder, line 1213: Does this need to return if the outputs are assigned to self?

docs/user_guide/methodology/qtbm.md
```diff
@@ -15,25 +15,37 @@ of the cost of a full NEGF calculation.
...
+always relies on the [`spectral` OBC algorithm](obc.md#spectral-method)
+to compute the boundary self-energies, since the injection vectors
+directly derive from the eigenpairs of the polynomial eigenvalue problem
+(see [NEVP](nevp.md)) that is solved in this OBC algorithm.
+
+If a contact's periodic cell is repeated $n_y \times n_z$ times in the
+two transverse directions, the boundary self-energy and injection
+vectors of the small unit cell are computed on a Monkhorst-Pack grid of
+transverse wavevectors and Bloch-summed back up to the size of the full
+contact block before being inserted into the device-sized system matrix.
```
vetschn, line 48: Maybe worth also linking to the SISPAD paper here once those proceedings are out 👍

src/quatrex/core/config.py
vetschn: The mid-gap energy of a reference contact is needed to compute the excess charge. The PR changes to output excess charge also in non SCSP runs

examples/cp2k/carbon-chain/qtbm-low-rank/quatrex_config.toml
vetschn, line 1: The Fermi levels are moved because they were nonsense for all devices before (all of them at -3 somehow) i guess, right?

src/quatrex/core/config.py
```diff
@@ -168,6 +168,10 @@ class QTBMConfig(BaseModel):
 
     """
 
+    atom_resolved_outputs: bool = False
+    """Whether to output atomic-resolved observables instead of 
+    orbital-resolved observables"""
```
vetschn, line 173: very good option! This only affects contact DOS, and excess charge densities?

src/quatrex/core/config.py
vetschn, line 1: Now setting a mid-gap energy for a grounded (reference) contact is no longer optional in QTBM runs. Not setting any causes a TypeError at runtime. Probably better to check this in the top-level config validators before starting the simulation

src/quatrex/core/qtbm.py
vetschn: The get_charge_density method needs this. Instead of assigning to an observables attribute it may be better to only return the arrays and set them where the method is called

src/quatrex/core/qtbm.py
vetschn: Not sure about the behaviour of np.save on cupy arrays, need to check this once alps is back

vetschn (commented):
Nice! I can see that this can be very useful. Left some comments here and there, it would be good to check the `reduceat` and whether a config validator for the mid-gap setting would make sense

Looking forward we should also parse through the outputs config to only have selected outputs in QTBM runs

src/quatrex/core/config.py
almaeder: Ah okay, at the contact refactor we need to discuss again about the combination of inputs required and the verification of it.

vetschn: No problem, i think the contact definitions are actually a super confusing part of the method, boggles my mind every time

examples/cp2k/carbon-chain/qtbm-low-rank/quatrex_config.toml
maurodossena: Yes, they were moved to make them consistent and obtain reasonable excess charges

src/quatrex/core/config.py
maurodossena: Yes, it only affects them.

src/quatrex/core/config.py
maurodossena: I added a check in the config validator.

src/quatrex/core/qtbm.py
maurodossena: Moved the reduction to the device. `np.save` should work with cupy array and move the array to host automatically.

src/quatrex/core/qtbm.py
maurodossena: Changed the code to assign current and excess charge ouside of the _compute functions.

maurodossena:
cscs-ci run

src/quatrex/core/config.py
```diff
@@ -2297,16 +2301,9 @@ def check_either_fermi_or_midgap(self) -> Self:
...
-            if (
-                contact.fermi_level is not None
-                and contact.mid_gap_energy is not None
-                and not (self.electron.band_edge_tracking or self.scsp is not None)
-            ):
+            if contact.mid_gap_energy is None and self.formalism == "wf":
                 raise ValueError(
-                    "Both `fermi_level` and `mid_gap_energy` cannot be set "
-                    "simultaneously, unless band edge tracking is active "
-                    "or the Schrödinger-Poisson solver is enabled."
```
vetschn, line 2309: Just to keep in mind: When we make the QTBM outputs configurable, this no longer needs to be enforced

vetschn (approved):

PR #341, opened 2026-08-20
Refactor `find_renormalized_eigenvalues`.

### LINKED ISSUE #302: Refactor find_renormalized_eigenvalues() to Handle Single Contact only and Remove Block Comm Broadcast
Refactor the `find_renormalized_eigenvalues()` method so that it only performs either the left or right contact calculation, rather than both. The goal is to eliminate the need to broadcast through the block communicator during this process, and simplify the control flow.

- Review the logic in `find_renormalized_eigenvalues()` to split responsibilities for left and right contact cases.
- Update the code to avoid broadcasting through the block communicator (`block comm`).
- Adjust the mechanism for printing diagnostic or status information to the terminal. You will likely need to update the rank (process) responsible for outputting info, as it may differ depending on the contact being computed.

(issue created with 🤖 )

almaeder:
This PR addresses issue #302.

It does the following:
- Removes duplicated code for left and right contacts in `find_renormalized_eigenvalues`.
- Changes the electron solver such that only ranks owning the contacts have their information.
- Removes unnecessary broadcasts through the block comm.

### DIFF
--- a/src/quatrex/electron/solver.py
+++ b/src/quatrex/electron/solver.py
@@ -150,75 +150,69 @@ class ElectronSolver(SubsystemSolver):
             for species in atomic_species
         ]
         orbital_coordinates = np.repeat(atom_coordinates, orbitals_per_atom, axis=0)
-        left_band_edge_info = xp.empty(3, dtype=float)
+
         if comm.block.rank == 0:
             # Quantities related to the left contact.
-            left_band_edge_info = self._configure_contact_band_edges(
+            (
+                self.left_fermi_level,
+                self.left_mid_gap_energy,
+                self.left_delta_fermi_level_conduction_band,
+            ) = self._configure_contact_band_edges(
                 config=config,
                 hamiltonian=self.hamiltonian,
                 overlap=self.overlap,
                 coordinates=orbital_coordinates[: self.block_sizes[0]],
                 side="left",
             )
+            self.left_voltage = config.electron.left_contact.voltage
+            self.left_mid_gap_energy -= self.left_voltage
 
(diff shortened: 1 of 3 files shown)


almaeder:
cscs-ci run

src/quatrex/electron/solver.py
```diff
@@ -150,76 +150,70 @@ def __init__(
...
-        left_band_edge_info = xp.empty(3, dtype=float)
+
         if comm.block.rank == 0:
             # Quantities related to the left contact.
-            left_band_edge_info = self._configure_contact_band_edges(
+            (
+                self.left_fermi_level,
+                self.left_mid_gap_energy,
+                self.left_delta_fermi_level_conduction_band,
+            ) = self._configure_contact_band_edges(
```
vetschn: Not sure how i feel about rank-local attributes that may raise AttributeErrors on other `comm.block.rank`s but i don't see how this could be a problem really 🤷

vetschn (approved):
Very nice, left one small comment

src/quatrex/electron/solver.py
almaeder: Fair, but they are really rank local, right? Even in combination with Poisson?

almaeder: But I also guess a domain-distributed SCBA with Poisson is unexplored.

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

vetschn:
cscs-ci run

PR #342, opened 2026-08-21
Contact Refactor

almaeder:
This PR will refactor the contact class.
The main goal is to remove the slicing of the Hamiltonian and overlap matrix in the initialization.
Further optimizations are made by implementing `quatrex preprocess` to discover the contact
and plot the contact band structure before running a simulation.

This PR will also touch part of the device class to enable devices from unit cell.

As usual, this PR will do much more than it was intended to do.


almaeder:
@maurodossena @vetschn most needed and not needed changes are done.
Only things left that could change: Adding back the functionality to compute the fermi level during `run`, writing the pre-processing fermi level to the `quatrex_config`, and batching the different inner-kpoint dimensions together in `compute_boundary`.

We can discuss in IRL whether the changes are needed.

Edit: The first two changes are done, but the batching shouldn't be done. I would suggest waiting on the batching until we have a better solution than using normal SciPy CSR matrices.

almaeder:
cscs-ci run

src/qttools/boundary_conditions/obc/spectral.py
```diff
@@ -159,6 +156,64 @@ def _extract_subblocks(
 
         return tuple(blocks[n:-n]) if n > 0 else tuple(blocks)
 
+    @staticmethod
+    def _upscale_subblocks(
+        blocks: tuple[NDArray, ...],
+        block_sections: int,
+    ) -> tuple[NDArray, ...]:
```
almaeder: This feels like duplicated functionality.
To be checked.
Functionality in `inputs` seems very similar. 
Should be combined and moved to `qttools`.

src/quatrex/bandstructure/contact.py
```diff
@@ -206,3 +211,197 @@ def objective_function(fermi_level):
...
+    Parameters
+    ----------
+    contact : Contact
+        The contact object.
+    contact_config : ContactConfig
+        The contact configuration object.
+    device : Device
+        The device object.
+    device_config : DeviceConfig
+        The device configuration object.
```
almaeder: Shouldn't need to pass in the config objects separately since they will be uniquely assigned to `Device` and `Contact`.

src/quatrex/device/device.py
```diff
@@ -362,3 +346,61 @@ def _add_contacts(self):
...
+                (contact_config.fermi_level is None)
+                or (
+                    contact_config.delta_fermi_level_conduction_band is None
+                    and self.config.scsp is not None
+                )
+                or (
+                    contact_config.mid_gap_energy is None
+                    and self.config.scsp is not None
+                )
+            ):
```
almaeder, line 388: Here, I am not fully sure.

src/quatrex/device/inputs.py
```diff
@@ -156,23 +169,93 @@ def _construct_transport_cell(
...
-    return cell
+
+        rows.append(xp.concatenate(row, axis=-1))
+
+    return xp.concatenate(rows, axis=-2)
+
+
+def expand_circulant_cell(
+    matrix_dict: dict,
+    transport_cell_size: int,
```
almaeder: I guess these two functions should be moved to `qttools.toeplitz`.

tests/quatrex/pre_processing/test_pre_processing.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2024-2026 ETH Zurich and the authors of the quatrex package.
+
+"""Tests for the pre-processing step of the quatrex package."""
+
+from pathlib import Path
+from typing import Callable
```
vetschn, line 8: Deprecated alias, better to use collections.abc.Callable

src/quatrex/core/config.py
```diff
@@ -941,26 +941,54 @@ class ContactConfig(BaseModel):
...
+      [`lattice_vectors`](#lattice_vectors) parameters to determine
+      which orbitals belong to the contact.
+
+    !!! warning
+        This parameter is currently only used in the `"wf"` formalism.
+
+    !!! note
+        There will be a future `"external"` option that will allow the
+        user to specify the contact elements directly. This will allow
+        to further reduce the contact size in the system matrix.
```
vetschn: How can this possibly work? We always need the full coupling of a contact to the device anyway, no?

src/quatrex/core/config.py
```diff
@@ -941,26 +941,54 @@ class ContactConfig(BaseModel):
...
+      to determine which orbitals belong to the contact. This is useful
+      for simple systems where the contacts are just contiguous
+      orbitals. Currently, [`name`](#name) is used to determine the
+      order. Thus, only "left" and "right" contacts are supported.
+      `"slice"` is more flexible than `"from_unit"` and can be used for
+      systems that are not constructed from a unit cell.
+    - `"real_space"`: Finds the contact orbitals based on their
+      real-space positions. This uses the [`origin`](#origin) and
+      [`lattice_vectors`](#lattice_vectors) parameters to determine
+      which orbitals belong to the contact.
```
vetschn: Can this not be inferred from which settings are present?

src/quatrex/core/config.py
```diff
@@ -972,8 +1000,69 @@ class ContactConfig(BaseModel):
...
+    If the [`lattice_vectors`](#lattice_vectors) parameter is set, this
+    is automatically computed from the lattice vectors. Otherwise, it
+    can be set manually.
+
+    !!! warning
+        This parameter is currently only used in the `"wf"` formalism.
+
+    """
+
+    delta_fermi_level_conduction_band: float | None = None
```
vetschn: This looks like it does not belong in the config

src/quatrex/core/config.py
```diff
@@ -1034,11 +1123,118 @@ class ContactConfig(BaseModel):
...
+                if (
+                    not np.isclose(self.cell_volume, volume_from_lattice)
+                    and comm.rank == 0
+                ):
+                    warnings.warn(
+                        f"The provided `cell_volume` ({self.cell_volume})/n"
+                        f"does not match the volume computed from `lattice_vectors` ({volume_from_lattice})."
+                    )
+
+        return self
```
vetschn: I don't think it should be allowed to provide both

src/quatrex/core/config.py
```diff
@@ -2112,6 +2330,34 @@ def set_dsdbsparse(cls, value) -> DSDBSparse:
...
+
+    compute_fermi_level: bool = False
+    """Whether to compute the Fermi level of the contacts.
+
+    This will compute the Fermi level of the contacts based on the
+    doping and the mid-gap energy.
+
+    !!! Note
+        This will override the `fermi_level` parameter in the contact
+        configuration.
```
vetschn, line 2278: So user inputs are overriden in the input itself?

src/quatrex/core/config.py
```diff
@@ -2159,6 +2405,16 @@ class QuatrexConfig(BaseModel):
...
     # --- Directory paths ----------------------------------------------
     config_dir: Path
+    """The directory where the configuration file is located."""
+
+    config_file: Path | None = None
+    """The path to the configuration file.
+
+    This is expected to be `None` and will be set automatically when
+    loading the configuration from a file. This is only used when the
+    configuration should be modified in `quatrex pre-process`.
```
vetschn: I think it is a bit strange to store this in a user-facing config that could be modified. Is there not a better way to achieve something similar?

examples/cp2k/carbon-chain/qtbm-low-rank/quatrex_config.toml
```diff
@@ -16,14 +16,14 @@ kpoint_shift = [0, 0, 0]
     name = "left"
     origin = [0.65, 0, 0]
     lattice_vectors = [[2.57031, 0, 0], [0, 10, 0], [0, 0, 10]]
-    direction = "a"
+    transport_direction = "a"
```
vetschn, line 18: I think the overall transport direction and contact directions are a bit different. I would not change the name. The move of both to a/b/c is sensible though as they are more "lattice" directions rather than real-space. Although i think the sparsity structure detection in NEGF actually assumes a cartesian index, so something to watch out for

examples/w90/mos2/gw-kpoints-symmetric/quatrex_config.toml
```diff
@@ -34,11 +34,11 @@ coulomb_screening_density = true
 
 
 [electron] # =========================================================
-left_contact.name = "source"
+left_contact.name = "left"
```
vetschn, line 38: Honestly not a big fan of having to "name" the contacts left and right always. What about another parameter "side" to determine this?

src/qttools/boundary_conditions/lyapunov/doubling.py
```diff
@@ -46,10 +45,8 @@ def __call__(
 
         Parameters
         ----------
-        a : NDArray
-            The system matrix.
-        q : NDArray
-            The right-hand side matrix.
+        a_q : tuple[NDArray, NDArray]
+            The matrices a and q of the discrete-time Lyapunov equation.
```
vetschn, line 49: Why this change? seems a bit strange? If you want to pop the system into the solver could you not just `*unpack` the system when you call?

src/qttools/boundary_conditions/obc/obc.py
```diff
@@ -119,7 +113,7 @@ def _fixed_point_step(
         ----------
         boundary_system : tuple[NDArray, ...]
             The boundary system to solve.
-            It is expected to be a tuple (a_ii, a_ij, a_ji)
+            It is expected to be a tuple (a_ji, a_ii, a_ij)
```
vetschn, line 116: That is the sane way 👍

src/quatrex/core/config.py
```diff
@@ -972,8 +1000,69 @@ class ContactConfig(BaseModel):
...
+
+    The slice is defined by the first and last indices of the contact
+    region in the system matrix. The slice is assumed to be contiguous
+    and correctly sorted. The slice spans both the origin of the contact
+    and its periodic images in the transport direction.
+
+    !!! warning
+        This parameter is currently only used in the `"wf"` formalism.
+
+    """
```
vetschn: `num_orbitals` or `size` instead of this?

src/quatrex/core/config.py
```diff
@@ -972,8 +1000,69 @@ class ContactConfig(BaseModel):
...
+    !!! warning
+        The `sections` in transport direction is ignored when
+        [`contact_finder_method`](#contact_finder_method) is set to
+        `"from_unit"` since the periodicity in transport direction can
+        be automatically inferred.
+
+    !!! warning
+        This parameter is currently only used in the `"wf"` formalism.
+
+    """
```
vetschn: so the block found in a slice is cut in half and those are then two transport blocks? first index is always transport direction or is this re-sorted based on the transport direction? In the case where the contact is not defined from structural info, sectioning seems so complicated that i would not even try to use it 😅 

Also why does it say that it is used in the `"from_unit"` case when it is actually automatically inferred there?

vetschn (commented):
Great effort to better unify the contact definitions and super nice functionality with the pre-processing command. I left some comments and i think we should discuss again the ways of how a user would want to define contacts in a simulation in general

src/quatrex/cli/main.py
```diff
@@ -250,6 +251,51 @@ def run(
...
+            help="Force abort the entire MPI environment on an unhandled exception to prevent hanging processes.",
+        ),
+    ] = True,
+):
+    """Run pre-processing tasks for the provided configuration."""
+    try:
+        config = _resolve_config_path(config)
+
+        from quatrex.core.config import parse_config, setup_context
+        from quatrex.pre_processsing import pre_process as main
```
vetschn, line 286: The mesh command aborts if run on multiple ranks, preprocess does not, why the difference?

src/quatrex/core/config.py
vetschn: Also why does the block_sections parameter still exist?

src/qttools/boundary_conditions/obc/spectral.py
almaeder: It has been moved, but the duplication has not yet been removed.

src/quatrex/core/config.py
almaeder: Mathieu was suggesting that. @maurodossena mentioned it.
I am not fully sure. Thus, I removed the comment for now.

src/quatrex/core/config.py
almaeder: Is done now. Old `slice` setting is removed as discussed in IRL since it doesnt really have practical use.

src/quatrex/core/config.py
almaeder: Fixed.

src/quatrex/core/config.py
almaeder: It will overwrite the config file, but it first makes a backup.
If a backup already exists, it raises an error.

src/quatrex/core/config.py
almaeder: Make it a `PrivateAttr`?

examples/cp2k/carbon-chain/qtbm-low-rank/quatrex_config.toml
almaeder: Hmm, the direction of the contact is where the waves are injected, in a sense, where the transport happens.
Only `direction` is a bit uninformative to me. Also, how it is used in `Contact` is always `transport` vs `transverse`.

src/quatrex/core/config.py
almaeder: And then write it after parsing?
Do we produce public documentation for private attributes of the config?

src/quatrex/core/config.py
almaeder: Hmm, but it is needed for Poisson. Pre-processing would set it, or it is computed at runtime.
Otherwise we always would need bandstructure calculations during runtime.

examples/w90/mos2/gw-kpoints-symmetric/quatrex_config.toml
almaeder: Could be done. It just needs to be a unique and canonical key per contact.
I think it is nicer to bind to the name instead of having an extra parameter.

src/qttools/boundary_conditions/lyapunov/doubling.py
almaeder: Change that could be reverted. In the discussion with Mauro, he suggested it would be nicer to pass in `a_xx` always and not always unpack. The interface needed to change for all solvers to expect a packed system.
Not happy with that change TBH.

src/quatrex/core/config.py
almaeder: Discussed IRL, the setting is gone.

src/quatrex/core/config.py
almaeder: Discussed IRL, the settings are gone.

src/qttools/toeplitz/circulant.py
```diff
@@ -0,0 +1,307 @@
...
+        expected = xp.concatenate(shifted_blocks, axis=-1)
+        actual = a[..., i * block_size : (i + 1) * block_size, :]
+
+        if not xp.allclose(actual, expected):
+            return False
+
+    return True
+
+
+def check_nested_phi_circulant(
```
maurodossena: 🧑‍🍳

examples/w90/mos2/gw-kpoints-symmetric/quatrex_config.toml
vetschn: Discussed irl and now have a better view of why we need to deal with contacts like this. We should add a check though that all contacts have unique names and the qtbm outputs should probably use the full names rather than the initial, what do you think @maurodossena ?

src/quatrex/core/config.py
vetschn: I can live with adding a comment to the modified lines to make the provenance clear :)

src/quatrex/core/config.py
vetschn: Looking at this more closely it is actually a bit cooked anyway, since we need the config file's parent directory during validation (resolve output paths etc.) So let's just guard against it being present in the toml and just set the absolute path to the config file during validation

examples/cp2k/carbon-chain/qtbm-low-rank/quatrex_config.toml
vetschn: Then let's leave it at `transport_direction` :+1:

src/qttools/toeplitz/circulant.py
```diff
@@ -0,0 +1,307 @@
...
+    n : int
+        The number of sections in the block circulant structure.
+
+    Returns
+    -------
+    int
+        The canonical index difference modulo `n`.
+
+    """
+    r = diff % n
```
maurodossena, line 27: Are there cases where diff >= n?

src/qttools/toeplitz/circulant.py
```diff
@@ -0,0 +1,307 @@
...
+    if r > n // 2:
+        r -= n
+    return r
+
+
+def _upscale_single_dimension(
```
maurodossena, line 33: I agree!

src/qttools/toeplitz/circulant.py
```diff
@@ -0,0 +1,307 @@
...
+    if block_index not in [-1, 0, 1]:
+        raise ValueError(f"Index must be -1, 0, or 1. Got {block_index}.")
+
+    # upscale first in transverse directions
+    matrix_dict = expand_transverse(
+        matrix_dict=matrix_dict,
+        transport_ind=transport_ind,
+        sections=sections,
+        phases=phases,
+    )
```
maurodossena, line 216: Is `expand_transverse` independent from `key_assumption`?

src/qttools/toeplitz/toeplitz.py
```diff
@@ -228,3 +237,223 @@ def expand_periodic_superblocks(
...
+
+    This function targets the case of real space matrices which are
+    toeplitz in the transport direction.
+
+    Parameters
+    ----------
+    matrix_dict : dict
+        The dictionary of matrices corresponding to different periodic
+        repetitions. It is assumed that only the upper parts are
+        present.
```
maurodossena: Which upper part? Because the default for `key_assumption` is None.

src/quatrex/bandstructure/contact.py
```diff
@@ -206,3 +210,194 @@ def objective_function(fermi_level):
...
+            phases=phases,
+            key_assumption="half",
+        )
+        for block_index in [-1, 0, 1]
+    )
+
+    return contact_band_structure(kpoints_transport, H_XX, S_XX)
+
+
+def compute_contact_band_properties(
```
maurodossena: Would it make sense to restrict the eigenvalues computation to a specific energy range? In case of big contact blocks with a lot of eigenmodes.

src/quatrex/bandstructure/contact.py
```diff
@@ -206,3 +210,195 @@ def objective_function(fermi_level):
...
+            transport_ind=contact.transport_direction,
+            block_index=block_index,
+            sections=contact.transverse_repetition_grid,
+            phases=phases,
+            key_assumption="half",
+        )
+        for block_index in [-1, 0, 1]
+    )
+
+    return contact_band_structure(kpoints_transport, H_XX, S_XX)
```
maurodossena: Would it make sense to restrict the eigenvalues computation to a specific energy range? In case of big contact blocks with a lot of eigenmodes.

src/quatrex/bandstructure/contact.py
```diff
@@ -206,3 +210,195 @@ def objective_function(fermi_level):
...
+    s_xx = s_xx_tmp
+
+    phases = tuple(np.exp(2j * np.pi * k) for k in kpoint)
+    phases = (
+        phases[: contact.transport_direction]
+        + phases[contact.transport_direction + 1 :]
+    )
+
+    H_XX = tuple(
+        construct_circulant_cell(
```
maurodossena: Another possibility would be to solve `ny` x `nz` k-dependent smaller problems and then concatenate the eigenvalues, similar to what we do for the OBCs. It would speed up the contact bandstructure calculations when dealing with big contacts.

src/quatrex/core/qtbm.py
```diff
@@ -569,16 +566,58 @@ def _assemble_system_matrix(
...
-        self._add_matrix_to_system_matrix(kpoint, -1, type="hamiltonian")
-        self._add_matrix_to_system_matrix(kpoint, energy, type="overlap")
-
-        if self.config.qtbm.low_rank_obc:
-            # No need to add the OBC self-energy to the system matrix
-            return
+        # NOTE: Another memory optimization could be to keep the k-point
+        # matrices in memory and not the real-spaces ones. This would
+        # only be more efficient if there are less k-points than real
+        # space points.
```
maurodossena, line 577: Good point! We could discuss this option. We should take into account though that real-space hopping hamiltonians are usually sparser than the k-points summed ones.

src/quatrex/device/contact.py
```diff
@@ -195,113 +204,100 @@ class Contact:
...
-
-        if self.origin_num_orbitals == 0:
-            raise ValueError(
-                f"Error in contact {self.name}: No atoms found inside the origin cell."
+        self.transport_direction = "abc".index(contact_config.transport_direction)
+
+        if contact_config._contact_finder_method == "real_space":
+            self.unit_cell_orbital_indices, repetition_grid, self.origin_key = (
+                real_space_discovery(
+                    hamiltonian=device.hamiltonians[0, 0, 0],
```
maurodossena, line 270: I'm still wondering if we might have to check all the hamiltonian coupings or if the (0,0,0) is enough. Need to discuss.

src/quatrex/device/contact.py
```diff
@@ -1223,18 +504,13 @@ def get_coupling_matrix(
         indices = self.transverse_to_transport_indices
         return coupling_matrix[indices.T, indices]
 
-    def _construct_contact_matrix(self, UC_matrix: dict, ky: float, kz: float):
+    def _construct_contact_matrix(self, UC_matrix: list):
```
maurodossena, line 703: Is this another toepliz matrix construction?

src/quatrex/device/contact_discovery.py
```diff
@@ -0,0 +1,810 @@
...
+
+    repetition_grid = list(transverse_repetition_grid)
+    repetition_grid.insert(transport_direction, num_transport_cells)
+
+    origin_key = (0, *origin_cell_offset)
+
+    return unit_cell_orbital_indices, tuple(repetition_grid), origin_key
+
+
+def simplified_discovery(
```
maurodossena, line 701: Nice feature!

src/quatrex/device/device.py
```diff
@@ -358,7 +343,63 @@ def _add_contacts(self):
...
+        Hamiltonian.
+
+        """
+
+        # NOTE: This is a temporary solution to avoid circular imports.
+        from quatrex.bandstructure.contact import compute_contact_band_properties
+
+        # The Fermi level is always required while midgap energy and
+        # delta_fermi_level_conduction_band are only required if SCSP is
+        # used.
```
maurodossena, line 376: Midgap is also required if `wf` is selected

src/quatrex/grid/energies.py
```diff
@@ -70,8 +70,9 @@ def get_electron_energies(config: QuatrexConfig) -> NDArray:
                 f"Could not find electron energies file at {energies_path}. Please provide an energy window in the config."
             )
 
-    if not os.path.exists(config.output_dir):
-        os.mkdir(config.output_dir)
-    xp.save(config.output_dir / "electron_energies.npy", electron_energies)
+    if comm.rank == 0:
```
maurodossena, line 73: Oof 💀

src/quatrex/device/inputs.py
```diff
@@ -106,82 +107,13 @@ def create_coordinate_grid(
...
-
-
 def _expand_tight_binding_matrix(
     matrix_dict: dict,
     num_transport_cells: int,
     transport_ind: int,
     block_start: int | None = None,
     block_end: int | None = None,
-    periodic_shift: tuple = (0, 0, 0),
+    transverse_shift: tuple = (0, 0),
```
maurodossena, line 116: Docstring needs to be updated

src/quatrex/pre_processsing/contact_bandstructure.py
```diff
@@ -0,0 +1,284 @@
...
+                endpoint=False,
+            )
+
+            e_k = compute_contact_bandstructure(
+                h_xx=h_xx,
+                s_xx=s_xx,
+                kpoint=kpoint,
+                contact=contact,
+                kpoints_transport=kpoints_transport,
+            )
```
maurodossena, line 100: We could merge this functions inside the Contact class in the future. Similar to the `compute_contact_band_properties` function

maurodossena (approved):

maurodossena (approved):
Great work! I only have some comments. 
I also tested against a bigger gate leakage simulation and the results look consistent.

examples/w90/mos2/gw-kpoints-symmetric/quatrex_config.toml
maurodossena: Yes I agree on checking unique names and using full names for the outputs.

src/qttools/toeplitz/circulant.py
almaeder: Not in the way it is used.

src/qttools/toeplitz/circulant.py
almaeder: Good catch. It should still be correct for `None` and `half`, but not for `upper`.
Note: Here, the assumptions are about those in the transport direction.
This is not well explained and the docstring will be updated..

src/quatrex/device/device.py
almaeder: Thx. Updated the qtbm from unit cell example.

src/quatrex/device/inputs.py
almaeder: Good catch. Thank you.

src/quatrex/pre_processsing/contact_bandstructure.py
almaeder: Hmm, I would keep the contact lean and keep these heavy functions separate, but we should discuss how to structure these parts.

almaeder:
> Great work! I only have some comments. I also tested against a bigger gate leakage simulation and the results look consistent.

Thank you very much. Doing a detailed test is best after so many changes.

src/quatrex/device/contact.py
almaeder: Here, I am not fully sure since it is used where the Fourier transform happens.

almaeder: I would keep this for now and revisit when batching this part.

src/qttools/toeplitz/toeplitz.py
almaeder: Another good catch. Too many wrong docstrings...

src/quatrex/bandstructure/contact.py
almaeder: It can make sense, but then we need to integrate shift-inverse.

src/quatrex/bandstructure/contact.py
almaeder: Makes sense. Should be part of a separate PR.

src/quatrex/device/contact.py
almaeder: We should probably pass in the maximal graph, but we probably will not change it until in leads to a bug.

almaeder:
@vetschn @maurodossena a last point I am not sure about: Should we really use the `cell_volume` of the smallest cell?

Also, there are still some valid comments, but I feel we should make some follow up issues with them.

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

almaeder:
cscs-ci run

src/quatrex/bandstructure/contact.py
vetschn: Quick note that this is actually also a difference compared with OMEN. In OMEN an Arnoldi solver is used to obtain the ~10-20 modes closest to mid-gap energy and that set of bands is also used to determine the Fermi level (which is a very reasonable approximation). Definitely something that we could also look into.

examples/w90/carbon-nanotube/qtbm-unit-cell/inputs/hamiltonian.h5
vetschn, line 1: Are the carbon nanotube unit cell inputs now duplicated for gw unit cell and for this as well?

src/quatrex/core/qtbm.py
vetschn: Yep, something we have discussed before, right? Should we open an issue for this to discuss a strategy?

src/quatrex/pre_processsing/contact_bandstructure.py
vetschn: I agree with @maurodossena here, this seems like something that could be part of the contact object. I don't think it's necessary to change this right now though.

src/quatrex/pre_processsing/contact_bandstructure.py
```diff
@@ -0,0 +1,286 @@
...
+        )
+        # NOTE: Naively take +-1eV around the center energy for the plot window.
+        plot_window = (center_energy - 1, center_energy + 1)
+    else:
+        plot_window = config.pre_process.plot_window
+
+    plt.ylim(plot_window)
+    plt.tight_layout()
+    plt.savefig(config.output_dir / "band_structure.png", dpi=300)
+    plt.close()
```
vetschn: Here you can instead call the methods directly on the actual axis and figure handles instead of using `plt.`. You should also call `matplotlib.use("Agg")` at the top to ensure it does not try to open an interactive window (this actually fails on my machine like it is now)

src/quatrex/pre_processsing/contact_fermi_level.py
```diff
@@ -0,0 +1,134 @@
...
+        # posixpaths
+        # Pop the config_file key since it is inferred
+        new_config.pop("config_file", None)
+        QuatrexConfig(**_parse_config(config_file, new_config.unwrap()))
+        config_file.write_text(tomlkit.dumps(new_config))
+
+    else:
+        warnings.warn(
+            "Automatic Fermi level computation does nothing for NEGF simulations"
+        )
```
vetschn, line 129: Why? Is it just not done here or what is the issue?

vetschn (commented):
I honestly cannot get a full grasp over the changes here anymore, but i trust that things still work 👍

almaeder:
> I honestly cannot get a full grasp over the changes here anymore, but i trust that things still work 👍

:skull: I wish I didn't feel like this as well.

examples/w90/carbon-nanotube/qtbm-unit-cell/inputs/hamiltonian.h5
almaeder: Yes, maybe it should be restructured to
`carbon-nanotube` and `carbon-nanotube-unit-cell`?
Or should all carbon nanotube examples be together even if they are quite different?

src/quatrex/core/config.py
vetschn: Where does this now happen? I cannot see it in the toml merge function

examples/w90/carbon-nanotube/qtbm-unit-cell/inputs/hamiltonian.h5
vetschn: or have an inputs-unit-cell directory next to the inputs?

src/quatrex/core/config.py
almaeder:
```
patch = dedent(
    f"""
    [[device.contacts]]
    name = "{contact.name}"
    {fermi_level=} # Set by `quatrex pre-process`
    {mid_gap_energy=} # Set by `quatrex pre-process`
    {conduction_band_edge=} # Set by `quatrex pre-process`
    """
)
```
The patch contains the comments.

examples/cp2k/carbon-chain/qtbm/quatrex_config.toml
vetschn, line 1: What would happen if i set different transport directions for device and contact? I guess in this specific example the device.transport_direction does not do anything. Maybe guard against setting both?

examples/w90/carbon-nanotube/inputs/device.msh
vetschn, line 1: Why did this change? Is it just because of the region names that were modified?

vetschn: ah probably, "left" is exactly two chars less than "source"

vetschn:
> @vetschn @maurodossena a last point I am not sure about: Should we really use the cell_volume of the smallest cell?

@almaeder I believe the cell volume should be that of the cell that the contact band structure is computed on. That should be the multiplied by the repetitions in transverse directions and the number of cells that are taken together in transport direction.

src/quatrex/pre_processsing/contact_bandstructure.py
vetschn: Actually after having a look again at the `bandstructure/contact.py` module i think it may be worth moving these methods already. I was very confused seeing `compute_contact_band_structure` and `contact_band_structure` in the same file. Maybe better to have both `compute_contact_band_structure` and `compute_contact_band_properties` as methods of the `Device` class?

src/qttools/toeplitz/circulant.py
```diff
@@ -0,0 +1,315 @@
...
+
+    for i in range(sections[1]):
+        for j in range(sections[1]):
+            block_size = a.shape[-1] // sections[1]
+            slice_i = slice(i * block_size, (i + 1) * block_size)
+            slice_j = slice(j * block_size, (j + 1) * block_size)
+            if not check_phi_circulant(a[slice_i, slice_j], sections[0], phases[0]):
+                return False
+
+    return True
```
vetschn: Could these two methods be moved to `test_circulant.py`?

vetschn:
Some notes i took while going through the changes again:

- The contact names "left"/"right" now determine the direction in which the contacts are connected to the scattering region in the case that the "contact finder method" resolves to "from_unit", which is always the case for NEGF runs and for QTBM it is the case only when the device is constructed from the unit cell hopping blocks.
- The contact's "direction" is now renamed to "transport_direction" and the axes are now a/b/c. This parameter is automatically checked and overridden by the contact config validators if it is not set in QTBM runs.
- Batching multiple energies together when computing OBC is no longer an option. This is a direct consequence of the contact not storing its own copies of the hamiltonian and overlap matrices. The `qtbm.max_batch_size` parameter is still there and by default set to `10` (some examples explicitly set it to `1`) but this does not actually have any effect.
- The tune_nevp.py script is deleted.
- The spectral OBC solver takes a list of block coefficients of the contact NEVP, it does not have to be exactly three. Upscaling and extraction of the authoritative boundary layers are done by methods in the `toeplitz` module.
- Structural contact discovery has been factored out into a separate contact discovery module. This module also implements the simplified discovery when upscaling the device from a unit cell.
- A new preprocessing step can plot contact band structure and Fermi level before a run and update them in the config file using `tomlkit`. If it is not set when starting a run, the Fermi level is still computed at runtime but a warning is raised.

src/quatrex/pre_processsing/contact_fermi_level.py
almaeder: Just not done yet, since I didn't think it would be worth it, since the `QTBM` code will become the final code.

examples/cp2k/carbon-chain/qtbm/quatrex_config.toml
almaeder: Fair, we only guard in one way and not the other. 
Is there a place where a global transport direction could be used while the contacts have other directions?
I do not believe so.

almaeder:
> > @vetschn @maurodossena a last point I am not sure about: Should we really use the cell_volume of the smallest cell?
> 
> @almaeder I believe the cell volume should be that of the cell that the contact band structure is computed on. That should be the multiplied by the repetitions in transverse directions and the number of cells that are taken together in transport direction.

Then we need to update the reference data since previously it used the smallest cell.

examples/cp2k/carbon-chain/qtbm/quatrex_config.toml
almaeder: Hmm, device transport direction is used for the k-points since we give them for all dimensions, not only `xyz`.
Also, it is used to drop inputs that have a non-zero value in the real-space index.
Currently, a global transport direction is needed.

almaeder:
cscs-ci run

almaeder:
> > > @vetschn @maurodossena a last point I am not sure about: Should we really use the cell_volume of the smallest cell?
> > 
> > 
> > @almaeder I believe the cell volume should be that of the cell that the contact band structure is computed on. That should be the multiplied by the repetitions in transverse directions and the number of cells that are taken together in transport direction.
> 
> Then we need to update the reference data since previously it used the smallest cell.

Maybe we do not, since the only example using it (`carbon-nanotube-poisson`) has no periodicity.

vetschn (approved):
🥇

almaeder:
cscs-ci run

maurodossena:
:shipit:

PR #343, opened 2026-08-21
Deformation potential electron-phonon scattering for 1D structures

### LINKED ISSUE #335: Long-wavelength phonons
A difficulty in implementing electron-phonon interactions is to obtain accurate coupling constants $\mathcal{M}_{\alpha \beta}$.
We should start by modeling long-wavelength phonons, where only the phonon modes and some deformation potential constants need to be known.

patkol:
This PR closes #335.

It adds a more sophisticated version of the "pseudo-scattering" that is implemented already. I called it "long-wavelength" because it should be accurate for long-wavelength phonons.

The self-energy is computed as 
$\Sigma^\gtrless(E)= \sum_{E_\mathrm{ph}} V_{E_\mathrm{ph}}^\mathrm{em} G^\gtrless(E\mp E_\mathrm{ph}) +  V_{E_\mathrm{ph}}^\mathrm{abs} G^\gtrless(E\pm E_\mathrm{ph})$, 
where the phonon energy $E_\mathrm{ph}$ is quantized according to the energy grid in use.
The prefactors $V_{E_\mathrm{ph}}^\mathrm{em}$ and $V_{E_\mathrm{ph}}^\mathrm{abs}$ are computed based on a given phonon dispersion `phonon_dispersion.npy` such that effectively
$\Sigma^\gtrless(E) = \frac{1}{N} \sum_{\vec{q}, \lambda} \left|M_{\vec{q} \lambda}\right|^2 \left( (n_\mathrm{p}(\hbar \omega_{\vec{q}, \lambda}) + 1) G^\gtrless(E\mp \hbar \omega_{\vec{q}, \lambda}) + n_\mathrm{p}(\hbar \omega_{\vec{q}, \lambda}) G^\gtrless(E\pm\hbar \omega_{\vec{q}, \lambda}) \right)$
is approximated, with coupling constants $M_{\vec{q} \lambda}$ from the deformation potential approximation.

Right now, this only works with 1D structures (no k-points) and requires the basis to be (near-)orthonormal to be accurate.

This is my first PR, so please watch closely :)


src/quatrex/core/scba.py
```diff
@@ -35,6 +35,17 @@
 profiler = Profiler()
 
 
+def _max(a: NDArray):
```
patkol: Should we move this function (or a better alternative) to qttools?

almaeder: Yes, but also change its name.
`_max` is not too descriptive.

patkol: Sure. Where would you put something like that? @vetschn suggested `qttools/comm/comm.py` for example?

almaeder: Would make sense, or in another file inside `qttools/comm`. 
Depends on whether it should become a method of comm, but I do not think it should
since it doesn't mirror a standard MPI function.

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
```diff
@@ -0,0 +1,68 @@
...
+obc.algorithm = "spectral"
+obc.nevp_solver = "full"
+obc.block_sections = 1
+
+# Parameters for Beyn's method.
+obc.num_ref_iterations = 2
+obc.max_decay = 15
+obc.min_decay = 1e-3
+
+obc.memoizer.mode = "off"
```
almaeder, line 53: We should only specify the parameters really needed.
I know the other examples also specify too much sometimes.

src/quatrex/core/config.py
```diff
@@ -1430,10 +1430,9 @@ class PhononConfig(BaseModel):
     self-energy.
 
     !!! note
-        Currently, only the `"pseudo-scattering"` model / deformation
-        potential interaction is implemented, which does not produce
-        any self-energy matrix elements besides the diagonal ones.
-
+        Currently, only the `"pseudo-scattering"` and `"long-wavelength"`
+        models / deformation potential interactions are implemented, which
+        do not take this variable into account.
```
almaeder, line 1477: Your model takes into account the off-diagonal elements, right?
Then it should be affected by the cutoff radius.
Also, if we have multiple interactions turned on, I believe we take the max radius. Maybe it would make sense to just have a single parameter and not one per physics model @vetschn?

src/quatrex/core/config.py
```diff
@@ -1445,13 +1444,11 @@ class PhononConfig(BaseModel):
...
-    model: Literal["pseudo-scattering", "negf"] = "pseudo-scattering"
+    model: Literal["pseudo-scattering", "long-wavelength", "negf"] = "pseudo-scattering"
     r"""Which model to use for the electron-phonon interaction.
 
-    Currently, only a monochromatic `"pseudo-scattering"` model is
-    implemented.
-
-    In this model, the electron-phonon interaction is modeled as
+    In the monochromatic `"pseudo-scattering"` model, the electron-phonon interaction
+    is modeled as
```
almaeder: Maybe we should remove the placeholder "negf" since it is not yet implemented, and not sure how soon it will be.

src/quatrex/core/config.py
```diff
@@ -1463,6 +1460,24 @@ class PhononConfig(BaseModel):
...
+    In the `"long-wavelength"` model, the self-energy is computed as
+
+    $$
+    \Sigma^\gtrless(E)
+    = \sum_{E_\mathrm{ph}}
+    V_{E_\mathrm{ph}}^\mathrm{em}
+    G^\gtrless(E\mp E_\mathrm{ph})
+    + V_{E_\mathrm{ph}}^\mathrm{abs}
+    G^\gtrless(E\pm E_\mathrm{ph}),
+    $$
```
almaeder, line 1515: Both in the old model and the new model, it is not mentioned how the spatial effect (only diagonal/ off-diaognals). Maybe we should amend the formulas/text for this.

src/quatrex/core/config.py
```diff
@@ -1474,6 +1489,19 @@ class PhononConfig(BaseModel):
     temperature: PositiveFloat = 300.0  # K
     """The temperature of the system in Kelvin."""
 
+    # Long-wavelength phonons
+    acoustic_deformation_potential: FiniteFloat | None = None
+    """The deformation potential of the acoustic phonon modes in eV."""
+    optical_deformation_potential: FiniteFloat | None = None
+    """The deformation potential of the optical phonon modes in eV/Å."""
```
almaeder: Are they allowed to be negative?

src/quatrex/electron/sse_phonon.py
```diff
@@ -48,6 +61,91 @@ def __init__(
...
+        if config.phonon.model == "long-wavelength":
+            self._compute_fn = self._compute_long_wavelength
+            if electron_energies is None:
+                raise ValueError(
+                    "Electron energies must be provided for the long-wavelength model."
+                )
+
+            # Load phonon modes
+            with open(config.input_dir / "phonon_dispersion.npy", "rb") as f:
+                """
```
almaeder: Would it make sense to check in the config validation that this file exists if the model is turned on?

almaeder: We currently do not check this for all inputs, but for some.
Should be done for all.

almaeder (commented):
Looks very good.

src/quatrex/core/config.py
patkol: I see. I don't take it into account in the code I added explicitely because my self-energy is just a superposition of Green's functions that I assume to have a reasonable cutoff already. Should I cut them again in `sse_phonon.py` with the cutoff given here, or does this happen somewhere else in the code?

src/quatrex/core/config.py
patkol: Yes. I'm anyway not sure what this would be exactly, we do NEGF with the other models too?

src/quatrex/core/config.py
almaeder: I think we should just have one cutoff that we use when assembling the sparsity pattern.
You shouldn't need to take it into account for the self-energy. Indeed, the Green's function is already cut, which happens when it gets initialized in SCBA.

almaeder: Maybe we can use this opportunity to make a single cutoff radius, except if it would make sense to have certain interactions over different ranges.

src/quatrex/core/config.py
patkol: The deformation potentials say how the conduction band changes under deformations. In principle that could go in both directions as far as I understand. I would not disallow negative values.

src/quatrex/core/config.py
patkol: So around this "negf" model there is quite a bit of a skeleton that's pre-made for non-equilibrium phonons. For example, the PiPhonon class in phonon/polarization.py is only referenced there.
If I would remove the "negf" model, I should also remove this skeleton, that would be quite a bit of code.
An alternative would be to rename the "negf" model and leave it as it is otherwise.
@vetschn, what is your opinion on that? I guess you worked on the skeleton?

patkol: After a discussion with @vetschn I removed the "negf" skeleton in [8956ae4](https://github.com/quatrex/quatrex/pull/343/commits/8956ae4de99e5b9d508aed0b244c700ea34d5966).

src/quatrex/core/config.py
patkol: I now mention in the pseudo-scattering documentation that only the diagonal is computed (Added in [8956ae4](https://github.com/quatrex/quatrex/pull/343/commits/8956ae4de99e5b9d508aed0b244c700ea34d5966))

src/quatrex/electron/sse_phonon.py
patkol: I added a check for the phonon dispersion and opened issue #344 for the remaining checks.

src/quatrex/core/config.py
```diff
@@ -1474,6 +1490,19 @@ class PhononConfig(BaseModel):
...
+    atom_mass: NonNegativeFloat | None = None
+    """
+    The mass of a single atom in (eV s^2) / Å^2.
+    All atoms are assumed to have the same mass.
+    """
+    lattice_constant: PositiveFloat | None = None
+    """The size of the unit cell along the transport direction in Å."""
+
```
almaeder, line 1547: Can this not be inferred from the `.xyz`?

src/quatrex/core/config.py
```diff
@@ -1474,6 +1490,19 @@ class PhononConfig(BaseModel):
     temperature: PositiveFloat = 300.0  # K
     """The temperature of the system in Kelvin."""
 
+    # Long-wavelength phonons
+    acoustic_deformation_potential: FiniteFloat | None = None
+    """The deformation potential of the acoustic phonon modes in eV."""
+    optical_deformation_potential: FiniteFloat | None = None
+    """The deformation potential of the optical phonon modes in eV/Å."""
+    atom_mass: NonNegativeFloat | None = None
```
almaeder: Could also be inferred, but for wannier we also often pass in `X` for the type.

src/quatrex/core/scba.py
```diff
@@ -357,16 +364,18 @@ def __init__(self, config: QuatrexConfig) -> None:
...
-                self.sigma_phonon = SigmaPhonon(...)
-
-            elif self.config.phonon.model == "pseudo-scattering":
+            if self.config.phonon.model == "pseudo-scattering":
                 self.sigma_phonon = SigmaPhonon(config, self.electron_energies)
 
+            elif self.config.phonon.model == "long-wavelength":
+                assert (
+                    self.electron_solver.overlap is None
+                ), "The long-wavelength model is only implemented for an orthonormal basis."
```
almaeder: Better to `raise` instead of `assert`.

src/quatrex/electron/sse_phonon.py
```diff
@@ -4,11 +4,23 @@
...
 from qttools.datastructures import DSDBSparse
+from quatrex.core import constants
 from quatrex.core.config import QuatrexConfig
 from quatrex.core.sse import ScatteringSelfEnergy
 from quatrex.core.statistics import bose_einstein
 
 
+def _get_equal_spacing(a: NDArray):
+    """Asserts that `a` is equispaced and returns the spacing."""
+    assert len(a.shape) == 1
```
almaeder: Similar here. Better `raise`.

src/quatrex/core/scba.py
patkol: Done in [d778f51](https://github.com/quatrex/quatrex/pull/343/commits/d778f51dc3e2509be0fd2e442ed9b55912a13b16)

src/quatrex/electron/sse_phonon.py
patkol: Done in [d778f51](https://github.com/quatrex/quatrex/pull/343/commits/d778f51dc3e2509be0fd2e442ed9b55912a13b16)

src/quatrex/core/config.py
vetschn: I think the initial idea must have been to compute different interactions for different cutoffs and that the maximum cutoff kind of dictates what is needed from the electronic GF. Honestly a bit half-baked, since we only ever had the GW interaction and the diagonal phonon scattering.

@almaeder From a physics perspective it may be that a cutoff per interaction makes sense so you can converge them differently. Do you think there could be any computational advantage to doing certain interactions only on a subset of the nnz? I guess there is a bit less computation in the FFT part but that should be negligible i guess :shrug:

src/quatrex/core/config.py
vetschn: I think as @patkol explained to me once the lattice constant is the one that was used to obtain the phonon dispersion, which can be something different from the transport cell

src/quatrex/core/config.py
vetschn: @patkol What about if the lattice is not monoatomic?

src/quatrex/core/config.py
patkol: Yes, `lattice_constant` is only used to allow `SigmaPhonon`'s `__init__` to figure out the phonon momenta. I think it's a good feature that the phonon dispersion can be computed in another unit cell.

examples/w90/carbon-nanotube/inputs/phonon_dispersion.npy
vetschn, line 1: I guess the shape of the phonon dispersion is `(num_modes, num_q_points)`. What unit is this dispersion in? Do the momentum points range from $-\pi$ to $\pi$?

Could you add a section on how this dispersion was constructed to the README.md file in the carbon-nanotube example directory?

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
```diff
@@ -0,0 +1,67 @@
...
+
+[phonon]
+interaction_cutoff = 5.5  # Å
+model = "long-wavelength"
```
vetschn: Out of curiosity: Why is it called "long-wavelength" model?

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
```diff
@@ -0,0 +1,67 @@
...
+temperature = 300  # K
+# Deformation potentials are of order 1 eV / 1 eV/Å:
+# https://journals.aps.org/prb/abstract/10.1103/PhysRevB.104.195201#s3
+acoustic_deformation_potential = 1  # eV
+optical_deformation_potential = 1  # eV / Å
+# 1.99264688270e-26 kg / (1e-20 / e)
+atom_mass = 3.192572275274879e-25  # (eV s^2) / Å^2
```
vetschn: Would it not be more natural to provide the mass in kg or in atomic mass units (in this case 12)? Or as @almaeder suggested one could alternatively just infer the atomic mass from the kind of atom in the structure.xyz

src/quatrex/core/config.py
vetschn, line 1530: Maybe you could rename this acoustic deformation potential and use it for both the pseudo-scattering and the long-wavelength model.

I would also add some `!!!note` admonitions to say that the parameter is only used in this or that model.

src/quatrex/electron/sse_phonon.py
```diff
@@ -48,6 +64,92 @@ def __init__(
...
 
+        if config.phonon.model == "long-wavelength":
+            self._compute_fn = self._compute_long_wavelength
+            if electron_energies is None:
+                raise ValueError(
+                    "Electron energies must be provided for the long-wavelength model."
+                )
+
+            # Load phonon modes
+            with open(config.input_dir / "phonon_dispersion.npy", "rb") as f:
```
vetschn: To load these files more efficiently in a distributed setting, you can use the `distributed_load` method from the `mpi_utils.py` module

src/quatrex/electron/sse_phonon.py
```diff
@@ -48,6 +64,92 @@ def __init__(
...
+                """
+                Specification on phonon_dispersion.npy:
+                This file contains the angular velocities `omega[mode, momentum]`
+                for the different phonon modes and momenta. The phonon momenta are
+                equally spaced as
+                `np.linspace(-pi/a, pi/a, n_phonon_momenta)`, with `a` the lattice
+                constant.
+                The longitudinal acoustic mode along x is the first one
+                (`omega[0, :]`), followed by the two transverse acoustic modes.
+                The remaining modes are in no particular order.
```
vetschn: Mentioned above already, this could go into the README.md in the example folder

src/quatrex/electron/sse_phonon.py
```diff
@@ -48,6 +64,92 @@ def __init__(
...
+                        * phonon_momenta
+                        * longitudinal_epsilon_x
+                    )
+                else:
+                    # Optical phonons
+                    coupling_constants[mode_index, :] = (
+                        config.phonon.optical_deformation_potential * prefactor
+                    )
+
+            energy_spacing = _get_equal_spacing(electron_energies)
```
vetschn: I think this is a lot more safe than what we have been doing in other cases (`dE = energies[1] - energies[0]` 💀 )

src/quatrex/electron/sse_phonon.py
```diff
@@ -4,11 +4,29 @@
...
+    differences = xp.diff(a)
+    spacing = differences[0]
+
+    if not xp.allclose(differences, spacing):
+        raise ValueError("`a` is not equispaced.")
+
+    return spacing
+
+
 class SigmaPhonon(ScatteringSelfEnergy):
```
vetschn: Since now the body of the class actually handles two completely different cases i would split the two models into two different classes `SigmaPhononPseudoScattering(ScatteringSelfEnergy)` and `SigmaPhononLongWavelength(ScatteringSelfEnergy)`, import both in `scba.py` and instantiate based on the model there

vetschn (commented):
Very nice, the vast majority is very clear and looks great already.

Besides updating the carbon-nanotube example's README, could you add a few words about the model to the methodology docs as well (there is already a tiny section about phonon interactions there)? Then it would be great to also make sure that the rest of the docs stays consistent with this addition before merging in.

vetschn:
Just thought about another thing: Does it not depend on the transport direction which the transverse branches are? So if transport occurs along the "y"/"b" direction instead, would one need to skip the acoustic phonon modes `[0, 2]` instead of `[1, 2]`?

src/quatrex/electron/sse_phonon.py
```diff
@@ -48,6 +64,92 @@ def __init__(
...
+                        * config.phonon.acoustic_deformation_potential
+                        * prefactor
+                        * phonon_momenta
+                        * longitudinal_epsilon_x
+                    )
+                else:
+                    # Optical phonons
+                    coupling_constants[mode_index, :] = (
+                        config.phonon.optical_deformation_potential * prefactor
+                    )
```
vetschn: How many phonon modes and q-points do you expect for larger devices? If this `for` loop should become a problem, it could be vectorized straightforwardly (`coupling_constants[0] = ...` and then the rest `coupling_constants[1:] = ...` )

src/quatrex/core/config.py
almaeder: Computationally, there shouldn't be a problem. We take the max cutoff to allocate the data. We just need to always set interactions to zero outside of the interaction-specific cutoff (I do not think we do this currently for GW).

Saving different NNZ per quantity would also be possible, but the distribution/implementation would be a bit cursed.

src/quatrex/core/config.py
almaeder: Ah okay, thanks.

src/quatrex/electron/sse_phonon.py
almaeder: :skull:

src/quatrex/electron/sse_phonon.py
```diff
@@ -114,3 +216,40 @@ def _compute_pseudo_scattering(
...
+            )
+            sigma_lesser.data[shift:, :] += (
+                self.V_abs[shift] * g_lesser.data[: ne - shift, :]
+            )
+            sigma_greater.data[shift:, :] += (
+                self.V_em[shift] * g_greater.data[: ne - shift, :]
+            )
+            sigma_greater.data[: ne - shift, :] += (
+                self.V_abs[shift] * g_greater.data[shift:, :]
+            )
```
vetschn: This loop appears to be quite slow also, but it could be replaced by a `sparse_banded_matrix @ data_vector` product over the energy axis that would probably be beneficial if the number of shifts is significantly less than the number of energies. The banded matrix would cost memory on the order of `num_shifts * num_energies`. Maybe @almaeder has a better idea?

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
patkol: Absolutely. @vetschn already cleaned the config briefly for me a while ago.
Which are the remaining unused parameters? I'm having trouble figuring out what's really not accessed and what would just fall back to some default.

src/quatrex/core/config.py
patkol: This sounds like a more general problem, should this go into a separate "refactor cutoffs" issue/PR?
Or should I just add some code now to cut specifically in the long-wavelength electron-phonon self-energy?

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
patkol: The deformation potential approximation is accurate for phonons close to the middle of the Brillouin zone, i.e., for long-wavelength ones.
In this regime it locally looks like the structure still has the same periodicity, but with slightly shifted atoms. The deformation potential models how these shifts change the electronic band structure.

I could also name it something like "deformation-potential", it's just potentially confusing that "pseudo-scattering" has a deformation potential too. Still, it might be a better name... After all, it's currently applied for short-wavelength phonons as well.

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
almaeder: The parameter for Beyn since Full is used.

src/quatrex/core/config.py
almaeder: Maybe better to have an issue and discuss it again.

src/quatrex/electron/sse_phonon.py
almaeder: Here, I would not aim for the best performance but for a clear reference implementation.

src/quatrex/core/config.py
patkol: @almaeder Does `X` mean unknown atom type? Or Wannier centers, not even atoms?

src/quatrex/core/config.py
patkol: As you commented below, I think it would be cleaner to fully separate the two models.

src/quatrex/core/config.py
almaeder: `X` means unknown, and we mostly use it for Wannier. So, it is not inferable there.

src/quatrex/core/config.py
almaeder: Would you suggest having the parameters for the different phonon models in separate tables?

src/quatrex/electron/sse_phonon.py
vetschn: Also fair, maybe a different picture also when run on GPU, which i have not tried yet. I had tried cranking up the number of energies to 1000 in the example, ran on 10 mpi ranks and then computing the interaction took more than twice the time that the whole ElectronSolver took

src/quatrex/electron/sse_phonon.py
patkol: I guess we could then make this a shared function. Same as for the `_max` function discussed above, I'm not sure where to put it...

src/quatrex/electron/sse_phonon.py
patkol: This is the main loop, so it would be great to speed it up, yes!

patkol: But I also agree with @almaeder , an optimization should not make it much more difficult to read in this case...

src/quatrex/core/config.py
patkol: @almaeder Do you mean something like `phonon.long_wavelength.acoustic_deformation_potential` vs `phonon.acoustic_deformation_potential`?

src/quatrex/core/config.py
patkol: @almaeder Okay, then I guess I'll leave the `atom_mass` in the config for now to keep it decoupled, similar to the discussion about the `lattice_constant`.

src/quatrex/core/config.py
almaeder: Potentially. @vetschn would you like this?

src/quatrex/core/config.py
almaeder: Both inferring and not monoatomic are open problems, but I would keep it like this for now and make an issue from the comment.

patkol: @vetschn I went back to the literature to check what would happen in the case of atoms with different masses. For acoustic phonons one should be able to use the arithmetic mean of the masses. For the optical ones I'm still not sure. However, this effectively just changes the deformation potentials, which we set arbitrarily right now anyway.
So I agree with @almaeder , I'll keep it as is for now and open an issue on the topic.

patkol: I opened issue #348 for this.

src/quatrex/core/config.py
patkol: I opened issue #349 for this.

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
patkol: So that's just `obc.num_ref_iterations`? Nothing changes if I remove it, but removing `obc.max_decay` leads to warnings
`quatrex/src/qttools/boundary_conditions/boundary_system.py:350: RuntimeWarning: High error at rank 0 for G: left-slice(0, 100, None) of OBCSolver Spectral:
  Relative recursion error: 1.436e-01
  Absolute recursion error: 6.625e+00`

src/quatrex/electron/sse_phonon.py
patkol: Yes, this makes it easier to read as well in my opinion. Implemented in 8a0b4a36.

src/quatrex/electron/sse_phonon.py
patkol: Implemented in 8a0b4a36

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
patkol: It would be more natural, absolutely. I didn't to it because I liked how the whole config file had consistent units (eV, s, Å) and I didn't want to break that.
I'll change this and put the unit into the variable name to avoid confusion.

src/quatrex/electron/sse_phonon.py
patkol: I must have screwed up in 8a0b4a3, the tests passed locally, but not on github :/

almaeder: Could be an issue from rebasing since we changed on dev to always use `abc` and not `xyz`.

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
```diff
@@ -0,0 +1,67 @@
+simulation_dir = "."
+input_dir = "../inputs"
+
+formalism = "negf" # "negf" or "wf"
+
+[device]
+transport_direction = 'x'
```
almaeder: This should be now `a`.

src/quatrex/electron/sse_phonon.py
patkol: Weird that this happened in this commit, I didn't rebase now?
But yes, looks like that's the problem...
So is my transport direction `"a"` now?

patkol: Ah yes, you mentioned it. Thanks.

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
patkol: My `config.py` still expects `["x", "y", "z"]`. Do the tests on github only work if I rebase to the current `dev`? Otherwise this seems very weird...

patkol: How can f761335 -> 8a0b4a3 with no rebase in between have caused this error?

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
patkol: Implemented in [3d13631](https://github.com/quatrex/quatrex/pull/343/commits/3d13631d411445a2e82cbfe0ce947aec92fd606f)

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
patkol: After a discussion with @vetschn, this really seems to be some kind of bug in how github tests. I'll just rebase and hope it works afterwards.

patkol: The rebase went through without any detected conflicts, so I guess github decided to rebase before testing. This behavior seems weird, can we turn this off somehow?
Anyway, I now rebased explicitely and fixed it in commit [eeb7ef4](https://github.com/quatrex/quatrex/pull/343/commits/eeb7ef4d8e4e217998699b933d34f3f94d7cbff1). Now the tests fail for even weirder reasons, I guess I'll just ignore them for now...

patkol:
> Just thought about another thing: Does it not depend on the transport direction which the transverse branches are? So if transport occurs along the "y"/"b" direction instead, would one need to skip the acoustic phonon modes `[0, 2]` instead of `[1, 2]`?

@vetschn Thank you for the comment. I thought a bit about this. I don't think we can require the ordering of the modes to be 
`0: acoustic along "a", 1: acoustic along "b", 2: acoustic along "c"`
For the carbon nanotube this would work because the transverse acoustic (TA) modes are exactly polarized along "b" and "c". But this is not always the case.
So I just require that mode index 0 is the longitudinal acoustic (LA) mode, and that the following two modes are the TA ones (for which the coupling vanishes). I'll clarify this in the documentation.
If we want to do transport along another direction, we should then be able to use the same phonon dispersion and use the same LA mode (so still index 0) as far as I understand. Keep in mind however that all this is for 1D structures only.

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
vetschn: Very weird behaviour, i think it could have something to do with the caching behavior of the `https://github.com/prefix-dev/setup-pixi` step in the CI. Opened an issue for this #350

src/quatrex/electron/sse_phonon.py
patkol: This is not specific to the example, this is the general specification for any 1D structure. I think it should be somewhere else, no?

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
patkol: I think I'll rename it to "deformation-potential".

patkol: Renamed in f330f19b.

src/quatrex/electron/sse_phonon.py
patkol: I benchmarked this again. On a CPU and with only 800 energies this loop takes 34% of the entire execution time. So I have to correct myself, I once said it was fast compared to the rest with 800 energies, don't know what I did differently then...
In that case I will really check whether I can speed it up.

almaeder: Hmm, the `A @ B` solution doesn't sound too efficient. Maybe we can write a simple kernel to decrease the number of global memory round trips.

almaeder: Something like loop over energy, then loop over shifts, and every thread does the 40 element reduction. It may not be the best, but it's better for cache coherence than the current NumPy version.

patkol: Okay, then I'll stop trying to optimize it further in python for now.

examples/w90/carbon-nanotube/electron-phonon/quatrex_config.toml
patkol: Removed `obc.num_ref_iterations` in 28d92550.

src/quatrex/electron/sse_phonon.py
patkol: `qttools/utils/solvers_utils.py`? Or a new module in `qttools/utils`?

almaeder: I would add another file in the comm folder.
We should start removing the `utils` module.

src/quatrex/electron/sse_phonon.py
patkol: After a discussion with @almaeder I'll continue working on the `sparse_banded_matrix @ data_vector` approach to hopefully end up at a maintainable and reasonably fast implementation. The memory footprint should not be too bad.

almaeder: I would imagine `dia` as a good format, but I believe cupyx/cusparse will always do the `spmm` in CSC.

src/quatrex/electron/sse_phonon.py
patkol: I can do that. What's the logic behind this - functions on NDArrays belong to `comm`? Because it does not explicitly have to do with communication, right?
Just asking so we don't start hiding the `utils` in random places, this would be worse than just having `utils` imo.

almaeder: Ah, the comment was for the `_max` function, which does the global maximum.
That one makes more sense to be in the comm module, but not be part of the communicator since it is not a standard mpi call.
The energy spacing could be put into `grid/energies`.

src/quatrex/electron/sse_phonon.py
patkol: I implemented the `sparse_banded_matrix @ data_vector` approach in [3915dd2](https://github.com/quatrex/quatrex/pull/343/commits/3915dd26101d00b2f92fc86befac665810b1ef39). It leads to a nice 6-7x speedup for the phonon self-energy computation with 800 energies on my laptop (CPU).

patkol: One could still try different sparse formats and so on, for example the format of `m_lesser` is not set explicitly as @almaeder pointed out. We can optimize further once we work on a large device.

src/quatrex/core/scba.py
patkol: Moved to `qttools/comm/utils.py` in [2722b4b](https://github.com/quatrex/quatrex/pull/343/commits/2722b4b7c460123ac0a8995e82b5d885c6510005). The corresponding test still needs to be sorted out, @almaeder is working on it.

almaeder:
cscs-ci run

src/quatrex/core/config.py
vetschn: I would not like this, i think it makes the config unnecessarily more complicated. Separating the two self-energy implementations makes sense because the code is quite different but one can still use the same parameter in the config

src/quatrex/electron/sse_phonon.py
vetschn: I think the README should still contain steps how to do create these inputs. Then the input_data.md section in the user guide can be renamed to "Input Data" instead of "Electronic Structure Data" and then you can put what is currently in the file under the heading "Electronic Structure" and add another heading about the phonon dispersion input.

src/quatrex/core/scba.py
patkol: @almaeder refactored the tests in [f49bfbe](https://github.com/quatrex/quatrex/pull/343/commits/f49bfbe599085f6be6a611e80ae6e381912a25e0) and I added another one in [e5bdc45](https://github.com/quatrex/quatrex/pull/343/commits/e5bdc4590f06b709d44eb25811df75b9bbef4bce).

src/quatrex/electron/sse_phonon.py
patkol: I added a shared function `get_equal_spacing` in [18fc7db](https://github.com/quatrex/quatrex/pull/343/commits/18fc7db77e61bddb1bf22938b8638b29f00d0c77) and applied it everywhere I could find the a[1] - a[0] pattern.

src/quatrex/electron/sse_phonon.py
patkol: Done in 1142371b. I think that made it much cleaner, thank you for the comment!

examples/w90/carbon-nanotube/inputs/phonon_dispersion.npy
patkol: The required input data is now (07dd3b22) specified in the documentation (User Guide -> Input Data -> Phonons).
And I describe how to calculate the dispersion from a force constants matrix in the README. The script I attached there is rather long, what do you think about that?

src/quatrex/electron/sse_phonon.py
patkol: I did that in 07dd3b22.

patkol:
> Very nice, the vast majority is very clear and looks great already.
> 
> Besides updating the carbon-nanotube example's README, could you add a few words about the model to the methodology docs as well (there is already a tiny section about phonon interactions there)? Then it would be great to also make sure that the rest of the docs stays consistent with this addition before merging in.

Yes, I updated the methodology docs in 07dd3b22.

src/quatrex/core/config.py
patkol: I ended up not adding separate tables, does this work for you @almaeder?

patkol: > Maybe you could rename this acoustic deformation potential and use it for both the pseudo-scattering and the long-wavelength model.
> 
> I would also add some `!!!note` admonitions to say that the parameter is only used in this or that model.

To come back to the original comment: I think I would prefer not to share the name because 
1. The `"deformation-potential"` model has two deformation potentials and it would be a bit arbitrary which one to share
2. The deformation potentials of the two approaches are not really comparable, they even have different units I think

src/qttools/comm/utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+    """
+
+    if xp.iscomplexobj(a):
+        raise ValueError("The maximum of a complex array is not defined")
+
+    # NOTE: numpy would return a scalar and thus, we use `atleast_1d`.
+    local_maximum = xp.atleast_1d(xp.max(a))
+    maximum = xp.empty_like(local_maximum)
+    comm.all_reduce(local_maximum, maximum, op="max")
```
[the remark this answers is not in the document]
patkol: @almaeder You already mentioned that the `distributed_max` function could be improved, should we do it now?

src/quatrex/electron/sse_phonon.py
```diff
@@ -114,3 +112,118 @@ def _compute_pseudo_scattering(
...
+        max_phonon_momentum = xp.pi / config.phonon.lattice_constant
+        phonon_momenta = xp.linspace(
+            -max_phonon_momentum, max_phonon_momentum, n_phonon_momenta
+        )
+
+        # Compute electron-phonon coupling constants
+        coupling_constants = xp.zeros((n_modes, n_phonon_momenta), dtype=complex)
+        atom_mass = config.phonon.atom_mass_u * constants.atomic_mass_unit
+        # prefactors[mode, qx]
+        prefactors = xp.sqrt(constants.hbar**2 / (2 * atom_mass * phonon_energies))
```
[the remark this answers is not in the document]
patkol: That's a good point - this can be seen as a problem with how the integration over phonon momenta is implemented. The integral has a pole at `omega == 0` which is not treated specially, and if we include this special point in the integration it explodes.

Since this is the canonical way to implement it and we don't need to optimize for coarse phonon momenta (everything that uses them is pre-computed) I would leave the code as is and document that `omega == 0` is not allowed. However, I now explicitly check for this edge case and raise an error if it occurs. I did this in commit [522a3b5](https://github.com/quatrex/quatrex/pull/343/commits/522a3b5a7b228440bedcd5f2a69098842f0d4248).

src/quatrex/electron/sse_phonon.py
```diff
@@ -2,58 +2,73 @@
...
+    """
+
+    if len(v.shape) != 1:
+        raise ValueError("`v` has multiple dimensions.")
+
+    v_size = xp.size(v)
+
+    diagonals = [xp.full(g_size - i, v[i]) for i in range(v_size)]
+    offsets = xp.arange(v_size)
+    m = sparse.diags(diagonals, offsets, format="csc")
```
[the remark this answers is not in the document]
patkol: This matters in the case there the matrix `m` occupies the full upper triangle. This is unlikely because phonon energies are typically smaller than the electronic ones, still good to catch this.

Issue #350, opened 2026-09-04
Make sure setup-pixi step does not cache a quatrex commit


vetschn:
The caching behavior of https://github.com/prefix-dev/setup-pixi needs to be checked, i think it might cache the editable install of quatrex itself as well leading to very strange behavior. Maybe also has to do with the `--frozen` setting?

PR #353, opened 2026-09-10
Bump version of github CI repo checkout action

vetschn:
Trying to reproduce the pixi caching bug from #350 in the linting CI here first. If it is indeed using the cached quatrex rather than the correct one, i'll change the `ci` pixi env not to include the repo by default.

### DIFF
--- a/.github/workflows/copilot-setup-steps.yaml
+++ b/.github/workflows/copilot-setup-steps.yaml
@@ -13,7 +13,7 @@ jobs:
         shell: bash -l {0}
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v7
+        uses: actions/checkout@v7.0.1
 
       - name: Set Up C++ Compiler
         run: |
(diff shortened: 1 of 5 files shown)


vetschn:
First run i got a cache miss, and the env was updated to the cache:
https://github.com/quatrex/quatrex/actions/runs/34449676903/job/102782320834?pr=353

The next time i got a cache hit but it still works as expected:
https://github.com/quatrex/quatrex/actions/runs/34449956266/job/102783199076?pr=353

@patkol I cannot reproduce the weird behavior that you saw. I guess it was something else strange? Pixi appears to install quatrex from the local folder using something that works like `pip install -e .` (which also makes sense from the lock file) so that's safe.

Maybe this was actually an issue with the checkout itself. Had a look at the issues there and maybe related to this? https://github.com/actions/checkout/issues/2511

I'll bump the checkout repo version to 7.0.1 where this should be fixed.

almaeder (approved):
LGTM, thank you for investigating.

patkol (approved):
Hard to be sure whether this fixes the issue if we cannot reproduce the problem, but let's try! Thank you for investigating this.

PR #354, opened 2026-09-15
Set up versioning for the documentation.

vetschn:
Added `mike` as a dependency and adjusted the docs workflow to allow tagging versions of the code. After this every time we merge something into `dev` the documentation will be updated and also every time we tag a release on `main`.

- Before merging this PR into `dev`, we have to delete what is currently on the `gh-pages` branch, otherwise it will bug. 
- The first time the workflow runs only on `dev`, the github pages deployment will 404 because no default version will be set (you can probably access the docs by appending `/dev` to the URL). Once merged into main, this fixes itself.

Tested this flow over at [github.com/vetschn/mike-test](https://github.com/vetschn/mike-test)


almaeder (approved):
LGTM

src/quatrex/core/config.py
```diff
@@ -1598,13 +1598,7 @@ def set_profiling_parameters(self) -> Self:
 
 
 class DeviceConfig(BaseModel):
-    """Configuration for the simulated device.
-
-    !!! warning
-        The contacts configuration in this table is only used in QTBM
```
almaeder, line 1604: upsli

